PR #4, opened 2018-07-11
Add '$' to args to make Add-Things work when fields have been filled in.

dizzi90:
Looks like a minor oversight rather than part of the exercise.


vexx32:
Sounds good, nice catch!

I'll merge it in.

Issue #12, opened 2018-07-13
Add comment based help for main module functions


vexx32:
Every Function Should Have Comment-Based Help!

vexx32:
Pretty much sorted for now, thanks Brandon! We can flesh it out as we go.

PR #14, opened 2018-07-13
Comment based help

ghost:
In reference to issue #12

### DIFF
--- a/PSKoans.psm1
+++ b/PSKoans.psm1
@@ -133,8 +133,8 @@ function Write-MeditationPrompt {
 
 	if ($PSCmdlet.ParameterSetName -eq 'Greeting') {
 		Write-Host -ForegroundColor Cyan @"
-    Welcome, seeker of enlightenment.
-    Please wait a moment while we examine your karma...
+Welcome, seeker of enlightenment.
+Please wait a moment while we examine your karma...
 
 "@
 		return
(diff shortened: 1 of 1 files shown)


vexx32:
This looks great! I have one or two comments though. One is that from the commit comparison it looks like it added an extra couple spaces to all the function lines when you put the help in. Not sure if that was intentional, but I've never seen it laid out like that.

Second thing is a parameter description or two needs tweaking:
```PowerShell
<# ...
.Parameter Clear
        Clear the screen before giving feedback. Defaults to True.
    .Parameter Meditate
        Opens your local koan folder.
    .Parameter Reset
        Resets everything in your local koan folder to a blank slate. Use with caution.
... #>
```
Also, the one in `Initialize-KoanDirectory` with the switch, that basically skips the prompt to confirm deletion / overwrite. (Reminds me, I should probably set it to prompt for confirmation by default, gotta raise that ConfirmImpact to High, or else just set `-Confim:$true` in the Get-Enlightenment call for it.)

Finally, looks like you use 2 space indents, and I'd prefer it we kept it consistent (everything else is at 4 spaces atm) 😄

BrandonLundt:
I believe I have the formatting resolved. And the parameter sections are updated.

One question on the Initialize-KoanDirectory. I'm really not clear on the purpose of FirstImport. If I call
`Initialize-KoanDirectory -FirstImport -Whatif`
I get the whatIf message AND copy-item still executes. Seems unusual/unexpected, as such not sure how to write the purpose of that parameter.

vexx32 (changes requested):
Lines 210 and 214 here remove the $script: prefix on $script:KoanFolder.

I actually don't know if that _breaks_ it, but I'd prefer it were explicit either way that the variable doesn't originate in the function itself. Especially with the if test keeping the prefix, it's unclear whether the two are different or not.

vexx32 (approved):

vexx32:
Looks good, I'll merge it in. 😄

Issue #13, opened 2018-07-13
Add actual Pester tests for module functions


vexx32:
Minus any that aren't... really... testable -- e.g., the standard output for the `rake` function is all Write-Host and shouldn't do anything. 
Perhaps that should be Mocked to just do nothing unless params are supplied? I'm not sure. Or at least the Write-MeditationPrompt should be mocked, anyway.
Anyway, need to figure all that stuff out to make sure the core module bits and pieces work properly!

BrandonLundt:
Regarding the tests, what does the build process look like? Is there a preference on psake vs invoke-build vs {Insert-BuildTool here} ?

vexx32:
Haven't got that straightened out yet, I'm open for suggestions. 😄

BrandonLundt:
I'm currently familiar with psake and have a template that we could use. It is loosely based on the psake build from Plaster. That said, I have been looking for an excuse to start using AppVeyor, so when this is ready for the PSGallery maybe we give that a go as well?

vexx32:
I'm definitely open to both, as I have limited experience with them and this is as good a time as any to learn. :)

vexx32:
Being worked on since #16

PR #16, opened 2018-07-14
Started root module testing

ghost:
In relation to #13 I have started unit testing for both Initialize-KoanDirectory and Get-Enlightenment. There is a lot more work to be done in this area, but wanted to get this submitted and see if any significant changes needed to be made.

### DIFF
--- a/Tests/PSKoans.tests.ps1
+++ b/Tests/PSKoans.tests.ps1
@@ -1,0 +1,120 @@
+$here = (Split-Path -Parent $MyInvocation.MyCommand.Path) -replace 'Tests', ''
+$sut = "PSKoans.psm1"
+Import-Module "$here\$sut" -Prefix "TST"
+
+Describe -Name "Initialize-KoanDirectory Unit Testing" -Fixture {
+    Context -Name "Mocked Functions: Test-Path,Remove-Item, and Copy-Item" -Tag @("Mocking") -Fixture {
+        Mock -CommandName Test-Path -MockWith { $false }
+        Mock -CommandName Remove-Item -MockWith {}
+        Mock -CommandName Copy-Item -MockWith {}
+
+        Initialize-TSTKoanDirectory -FirstImport
+
+        It -name "Assert Test-Path mocked" -test {
+            Assert-MockCalled -CommandName Test-Path -Times 1
+        }#It
+        It -name "Assert Remove-Item mocked" -test {
+            Assert-MockCalled -CommandName Remove-Item -Times 0
+        }#It
+        It -name "Assert Copy-Item mocked" -test {
+            Assert-MockCalled -CommandName Copy-Item -Times 1
+        }#It
+    }#Context Mocked Functions
+    Context -Name "Mocked Functions: Remove-Item, and Copy-Item" -Tag @("Mocking") -Fixture {
+        $KoanFolder = $Home | Join-Path -ChildPath 'PSKoans'
+        If( Test-Path $KoanFolder){
+            Remove-Item $KoanFolder -Recurse
+        }
+        Mock -CommandName Remove-Item -MockWith {}
+        Mock -CommandName Copy-Item -MockWith {}
+
+        Initialize-TSTKoanDirectory -FirstImport
+
+        It -name "Assert Remove-Item mocked" -test {
+            Assert-MockCalled -CommandName Remove-Item -Times 0
+        }#It
+        It -name "Assert Copy-Item mocked" -test {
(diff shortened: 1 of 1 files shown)


Tests/PSKoans.tests.ps1
```diff
@@ -0,0 +1,120 @@
...
+        Initialize-TSTKoanDirectory -FirstImport
+
+        It -name "Assert Remove-Item mocked" -test {
+            Assert-MockCalled -CommandName Remove-Item -Times 0
+        }#It
+        It -name "Assert Copy-Item mocked" -test {
+            Assert-MockCalled -CommandName Copy-Item -Times 1
+        }#It
+    }#Context Mocked Functions
+    Context -Name "No mocking, live execution" -Fixture {
```
vexx32: If we're doing a test with live execution, we should probably remove the created files after the test is complete. Also... we may want to build in some kind of 'debug mode' where it creates the koan directory in `$env:TEMP` instead, so that we don't overwrite anything that already exists when running tests.

Tests/PSKoans.tests.ps1
```diff
@@ -0,0 +1,120 @@
...
+    }#Context Mocked Functions
+    Context -Name "Mocked Functions: Remove-Item, and Copy-Item" -Tag @("Mocking") -Fixture {
+        $KoanFolder = $Home | Join-Path -ChildPath 'PSKoans'
+        If( Test-Path $KoanFolder){
+            Remove-Item $KoanFolder -Recurse
+        }
+        Mock -CommandName Remove-Item -MockWith {}
+        Mock -CommandName Copy-Item -MockWith {}
+
+        Initialize-TSTKoanDirectory -FirstImport
```
vexx32: Since this is not an exported command from the module... will it work to do this? Or will we need to call `Get-TSTEnlightenment -Reset` to trigger it instead?

Tests/PSKoans.tests.ps1
```diff
@@ -0,0 +1,120 @@
+$here = (Split-Path -Parent $MyInvocation.MyCommand.Path) -replace 'Tests', ''
+$sut = "PSKoans.psm1"
+Import-Module "$here\$sut" -Prefix "TST"
```
vexx32: Is the fact that `Initialize-KoanDirectory -FirstImport` effectively called when loading the module a concern here? Just trying to avoid having invoking tests create a koans directory / potentially overwrite it just for the sake of running tests.

It's not a huge concern now, but as we get more and more koan files... it might become more important later on.

Tests/PSKoans.tests.ps1
```diff
@@ -0,0 +1,120 @@
+$here = (Split-Path -Parent $MyInvocation.MyCommand.Path) -replace 'Tests', ''
```
vexx32: I'd suggest a more explicit variable name like `$KoanModulePath` or something along those lines.

Tests/PSKoans.tests.ps1
```diff
@@ -0,0 +1,120 @@
...
+$sut = "PSKoans.psm1"
```
vexx32: Suggest variable name be made more explicit - `$KoanModuleFile` perhaps?

Tests/PSKoans.tests.ps1
BrandonLundt: Yes this will work as we are importing the psm1, not the psd1. So at this point everything is exported.

Any value in later on testing what commands are exported? Maybe just a check to see that all exported commands exist in the psm1?

Tests/PSKoans.tests.ps1
BrandonLundt: Agreed, I like the idea of individual module files in the repo. Then the build process combines them into a single root module.

I will try to get around to that tonight.

Tests/PSKoans.tests.ps1
vexx32: Can't hurt to make sure that everything is present, absolutely!

Tests/PSKoans.tests.ps1
BrandonLundt: This follows the "outline" provided by new-fixture in the pester module. If we do end up keeping everything in one module file, I agree it would make sense to rename.

Tests/PSKoans.tests.ps1
BrandonLundt: I debated quite a bit when writing this test. I ultimately agree with your recommendations.

PR #26, opened 2018-07-16
Discrete testing

ghost:
As discussed during pull request #16 having the functions split into individual files would make testing simpler.  I also added a -Destination parameter to make it easier on the pester tests to generate test data at a separate location. 

If this looks good, will start work on the other tests we discussed. Namely only functions defined in the public folder are exported.


vexx32:
I'll have a proper looksee tomorrow, but for the moment -- I can't say I've ever seen NestedModules used in this fashion, nor have I seen many modules work like this. The closest I've seen is a primary module file that dot-sources the individual function files.

Is this better than that approach? Are there any potential scope issues that might crop up from using NestedModules? Will these nested modules show up as separate modules with `Get-Module`? I would think that would be... an undesirable consequence, as it might make it more complex to `Remove-Module`.

Just trying to see how this would play out and whether this particular approach to splitting the function files is the best way to go. I definitely agree it is something we'll need to do, but I'm not sure this is precisely how we want to go about it. :)

BrandonLundt:
> Is this better than that approach?
I have not tested speed for loading differences between nested and dotsourcing. That said, it is my opinion that this is a better approach on the basis of your manifest now has a clear definition of what files are being processed.

> Are there any potential scope issues that might crop up from using NestedModules?
I have used this method on both posh 4.0 and 5.0 and have not had any scoping issues. Anything older than 4.0 I have not tested.

> Will these nested modules show up as separate modules with Get-Module?
Negative. Get-Module returns the same data and the commands load like they were defined in the root module.

vexx32:
Alrighty then, sounds good!

At some point, it may be worth putting in a build task to recombine the disparate module files, but for now this ought to work perfectly well. :)

vexx32:
A concern I have with the -Destination parameter here is that if the user ever wanted to specify a different folder, they have to specify it every time they call `Get-Enlightenment`

There's that disconnect between them here. Maybe it'd be better for the -Destination parameter to function simply as a 'check against stored value, update if needed' and then have everything else still work with the stored value.

Perhaps having it stored in an environment variable or a data file would be best?

BrandonLundt:
Regarding indentation, are we going with tabs or spaces? That seems to be the source of my indentation issues. I'm good with either, I just want to set VSC to follow one or the other.

Regarding the here strings what do you think if this:
`Write-Host -ForegroundColor Cyan @"
Welcome, seeker of enlightenment.
Please wait a moment while we examine your karma...

"@`

Was like this: `$Message = "Welcome, seeker of enlightenment. `nPlease wait a moment while we examine your karma...`n"`
Write-Host -ForegroundColor Cyan -Object $Message`
Github formatting is eating my backtick on the n for a newline.

Regarding the -Destination parameter, you are correct this could be cumbersome for users if they choose a different path than the default. However if they use the default all functionality remains the same. Right now I'm leaning towards using an environment variable, But need to noodle this a little more.

vexx32:
I would... _prefer_... To avoid using escape sequences... But it might be a bit simpler. Oh, tell you what, here's a fun solution (I think I can roll it myself without much hassles. I'll define all the here strings in one hashtable and then just `Write-Host @Cyan $Table['Greeting']` etc. That will keep all the indentation weirdness to one place, so if it messes up, no big deal.

I have VS Code set up for spaces, 4 spaces to a 'tab', and I am a little partial to it. :smile:

Yeah I'm not sure where to go with -Destination yet either. If we go Env variable, it will have to be set in the system/user global Env vars to be persistent,, and I'd prefer to keep things mostly contained within the module. Maybe have it stored in a ps1xml file like the random zen quotes would be fine, and have it imported into the $script: or $env: scoped variable on module import.

vexx32:
@BrandonLundt 
So I figured out how to sort koans by using an attribute on the koan test files. You may have a bit of a tidying up to do before I can merge it, especially as I renamed a bunch of files -- sorry! The `Get-Enlightenment` function also got a small section modified, so make sure that gets carried over and split out properly. 

Let me know if you have any issues, I like the direction this is going. :D

BrandonLundt:
So, my attempts at resolving the merge conflicts have gone _very_ poorly. To the point of, I might delete my branch and start over. :)

What I have tried so far:

```
git pull upstream master --rebase
```

Merge conflicts!!!! YEAH!!

```
git add .\PSKoans.psm1
git add .\PSKoans\PSKoans.psd1
git rebase --continue
```

I then get : 

```
git status
On branch DiscreteTesting
Your branch and 'origin/DiscreteTesting' have diverged,
and have 103 and 15 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
```

So I says
```
git pull
fatal: Not possible to fast-forward, aborting.
```

I'm happy to start over on this branch or if you now how to resolve this "locally" I'm happy to do that as well.

vexx32:
To be honest... I don't really know how to resolve it... 😦 

I figured it might be a bit... tricky. It probably will be easiest to just save a copy of your test files, pull from the current dev and split things out from there. 

So I've also gone ahead and started getting a CI/CD pipeline going, which is... semi-function for now. Won't know for sure without some more attempts, but... we'll see. I've got it hooked into AppVeyor, anyhow, and got some stuff going.

So now's the time to get some tests going 😄 

Let me know if there are any issues or what have you, and keep in mind that I've made... probably too many... changes to how things work. Notably, I've renamed the $script scope variables; you'll see 'em. And... a bunch of other stuff... hehe. I have been trying to keep release notes more or less up to date, so feel free to browse those too ^^

vexx32:
I'll close this for now to keep things tidy; you just grab what you need and I'll try to avoid major changes to the structure till we get a good bit of this stuff in! 😄 

Feel free to open a new PR when you redo the splitting.

PR #36, opened 2018-08-02
Feature/stringoperatorfix

rdbartram:
fix couple of small issues with string foundational koans.
ps. love the work


vexx32:
Oh, nice catch! Look at me, completely forgetting how matching works. Thank you!

PR #38, opened 2018-08-03
Feature/loopfix

rdbartram:
I thought the string example was not clear since the element 0 was used twice.

Also in the loops, the values weren't being returned because they were using the ++syntax.
 Works when assigning the value but not for return to output....who knew!
 Fix is to use write-output.

Last point was to demonstrate use of ++$i vs $++

### DIFF
--- a/PSKoans/Koans/Foundations/AboutLoopsAndPipelines.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutLoopsAndPipelines.Koans.ps1
@@ -61,19 +61,32 @@ Describe 'Pipelines and Loops' {
         $Values = foreach ($Number in 1..5) {
             # In these kinds of loops, the specified variable name ($Number in this case) takes the place
             # of the $_ or $PSItem automatic variables instead.
-            $Number++
+            $Number
         }
         $Values | Should -Be __
 
         $Values = for ($i = 0; $i -lt 5; $i++) {
             # For loops are quite rare in native PowerShell code. They have their uses, but are
             # frequently too semantically obscure to have a place in PS's verbose ecosystem.
+            # Remember, ++ after the variable mean assign the value then increment.
+            # ++ before the variable means increment and the assign the value.
             $i
         }
         $Values | Should -Be @(0, 1, 2, 3, 4)
 
-        $Values = while ($true) { # watch out for infinite loops!
-            $Tick++ # Remember: an undeclared variable acts as zero until we increment it!
+        $Values = while ($true) {
+            # watch out for infinite loops!
+
+            # Remember: an undeclared variable acts as zero until we increment it!
+            # Incrementing a variable won't output the value
+            # unless the expression is wrapped in parentheses.
+            (++$Tick)
+
+            # An alternative would be the following:
+            # $returnvalue = ++$Tick
+            # $returnvalue
+            # Perhaps slightly more clear but more verbose
+
             if ($Tick -gt 2) {
                 break # the break statement breaks out of the current loop.
             }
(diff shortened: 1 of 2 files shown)


PSKoans/Koans/Foundations/AboutLoopsAndPipelines.Koans.ps1
```diff
@@ -61,19 +61,21 @@ Describe 'Pipelines and Loops' {
...
             # frequently too semantically obscure to have a place in PS's verbose ecosystem.
+            # Remember, ++ after the variable mean assign the value then increment.
+            # ++ before the variable means increment and the assign the value.
             $i
         }
         $Values | Should -Be @(0, 1, 2, 3, 4)
 
         $Values = while ($true) { # watch out for infinite loops!
-            $Tick++ # Remember: an undeclared variable acts as zero until we increment it!
+            Write-Output (++$Tick) # Remember: an undeclared variable acts as zero until we increment it!
```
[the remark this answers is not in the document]
rdbartram: I've updated the code as you suggested. I would be interested to know why you don't think Write-Output is best practice to be teaching.

PSKoans/Koans/Foundations/AboutLoopsAndPipelines.Koans.ps1
```diff
@@ -74,8 +74,19 @@ Describe 'Pipelines and Loops' {
...
         $Values | Should -Be @(0, 1, 2, 3, 4)
 
-        $Values = while ($true) { # watch out for infinite loops!
-            Write-Output (++$Tick) # Remember: an undeclared variable acts as zero until we increment it!
+        $Values = while ($true) {
+            # watch out for infinite loops!
+
+            # Remember: an undeclared variable acts as zero until we increment it!
+            # Also, to make sure it is returned the value must be wrapped in brackets
+            # else the variable is simply output and not returned.
```
vexx32: This is ambiguous; we want the variable value to be output, after all! To powershell, output and 'return' are often synonymous.

Perhaps something more like... "Incrementing or assigning a value to a variable won't output the value unless the expression is wrapped in parentheses." 

Or something along those lines!

rdbartram: facepalm moment. I mean't to write "else the variable is simply assigned and not output." but I was talking to someone at the time and wrote it wrong. I've added your description and squashed my commits. I don't know if you squash or not

vexx32: I don't github much, so that is gibberish to me. I'll read up on it later :)

PSKoans/Koans/Foundations/AboutLoopsAndPipelines.Koans.ps1
```diff
@@ -61,19 +61,32 @@ Describe 'Pipelines and Loops' {
...
 
-        $Values = while ($true) { # watch out for infinite loops!
-            $Tick++ # Remember: an undeclared variable acts as zero until we increment it!
+        $Values = while ($true) {
+            # watch out for infinite loops!
+
+            # Remember: an undeclared variable acts as zero until we increment it!
+            # Incrementing a variable won't output the value
+            # unless the expression is wrapped in parentheses.
+            (++$Tick)
```
vexx32: SO i just tried this one to be absolutely sure... and the parentheses _aren't_ necessary here, to my surprise:
```powershell
$a = $b++ # $a is 0, $b is 1
$c = ++$d # $c and $d are 1
```

PR #40, opened 2018-08-04
Add Stock Example Challenge

vexx32:
Borrowed from the F# koans. :)

Comments welcome, I'm still considering whether something like this makes sense here.


vexx32:
That's the idea! :P

PR #56, opened 2018-10-03
Add Clear-Path alias to Measure-Karma

steviecoaster:


vexx32:
Please add the alias to the .PSD1 as an `ExportedAliases` entry as well? 😄

PR #57, opened 2018-10-05
Create Documentation Foundation for Koans. 

romero126:
Update Koan Class Object to improve Documentation Capabilities


PSKoans/Public/Measure-Karma.ps1
```diff
@@ -90,6 +98,19 @@ function Measure-Karma {
                 Write-Verbose "Karma: $KoansPassed"
                 if ($PesterTests.FailedCount -gt 0) {
                     Write-Verbose "Your karma has been damaged."
+                    if ($ShowHelp) {
+                        $KoanAttribute = (Get-Command $KoanFile).ScriptBlock.Attributes | Where-Object TypeID -match 'KoanAttribute'
+                        Write-Host ($KoanAttribute | Out-String)
+                        if ($KoanAttribute.HelpURL) {
+                            Write-Verbose "Confusion is only the beginning."
```
vexx32: This should probably be a Write-Host.

PSKoans/Public/Measure-Karma.ps1
```diff
@@ -90,6 +98,19 @@ function Measure-Karma {
...
+                            start $KoanAttribute.HelpURL
+                        }
+                        if ($KoanAttribute.HelpAction) {
+
+                            Invoke-Command $KoanAttribute.HelpAction
```
vexx32: That's an interesting thought indeed, but I'd probably prefer to simply `&` this or `.InvokeReturnAsIs()` since it's a script block being executed locally.

vexx32: Also, it looks like you removed the HelpAction in the koan class. Either remove this or add it into the koan class. 😄

PR #58, opened 2018-10-07
More descriptive getting started

nohwnd:
Rewrote the getting started, because I had trouble grasping what I should do when starting out, and I already know about Pester and Powershell. The upper part in readme now describes what pskoans is in general, and links to the bottom where a full guide is. Personally I would also move the repo installation under the getting started, because almost everyone uses psgallery now, but did not want to change that much. I also changed the first koan, because the readme on the top of the file mentions that __ should be replaced, yet the first koan did not use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,18 +4,17 @@
 |     :---:     |      :---:      |
 | [![Build status](https://ci.appveyor.com/api/projects/status/komkhtba6v3n7iji/branch/master?svg=true)](https://ci.appveyor.com/project/vexx32/pskoans/branch/master) | [![Build status](https://ci.appveyor.com/api/projects/status/komkhtba6v3n7iji?svg=true)](https://ci.appveyor.com/project/vexx32/pskoans) |
 
-Inspired by Chris Marinos's **fantastic** [F# koans](https://github.com/ChrisMarinos/FSharpKoans), the goal of the PowerShell koans is to teach you PowerShell through Pester unit testing.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -37,16 +36,89 @@ Install-Module Pester -Scope CurrentUser
...
+2. Run `Measure-Karma -Meditate` to open your Koans folder and navigate to `Foundations\AboutAssertions.Koans.ps1`. Near the top you'll see:
+
+```powershell
+It 'is a simple comparison' {
+    # Some truths are absolute.
+    __ | Should -Be $true
+}
+```
+
+The `__` represents a blank for you to fill, and `| Should -Be $true` shows the expected result. To pass this koan you need to replace `__` with `$true`, like this: `$true | Should -Be $true`.
```
vexx32: I'm not entirely convinced we should have a complete walkthrough for the first koan. Sure, it's simple as heck, but it seems a little... weird. I dunno.

nohwnd: @vexx32 the koan is extremely simple, but understanding what you should do to take on this journey is not.

Imagine you don't know anything about powershell, pester, testing or koans. Then the first page you see offers no real info, because you don't know what Describe, It or Should is. You don't know that you are looking at a mix of standard PowerShell messages, Pester messages, and custom stuff.

So you try the second command which opens vscode without pointing you to any file. And then you just click around lost. Like I did this morning. 😄 

So if it took me good 10 minutes (and two visits to google) to understand what you want me to do, then for a beginner it won't be much easier. I admit I scrolled directly to getting started on the github readme, and if I did not do it I might have had easier time, but that's what people do, they skim stuff. Which is one of the reasons I included the whole screens with some colors, so they anchor your vision to something similar to what you see on the screen.

vexx32: Yeah, that's fair enough, haha.

Thank you for this! 😄

PR #109, opened 2018-12-15
Fix path issues on different platforms

Tyrix:
This should be os agnostic (core).

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Install-Module Pester -Scope CurrentUser
 1. `git clone` the repository into your desired directory, or download the repository as a .zip file and extract into a directory of your choice.
 2. Then from a normal powershell session run `Get-ChildItem -Recurse | Unblock-File` in that directory to remove the "downloaded from internet" flag that blocks them from running.
 3. Check `Get-ExecutionPolicy`: if it says 'Restricted' or 'Undefined', you need to also run `Set-ExecutionPolicy RemoteSigned` in order to allow the scripts to run.
-4. Run `Copy-Item -Recurse "$ProjectRoot/PSKoans" -Destination ($env:PSModulePath -split ';')[0]`
+4. Run `Copy-Item -Recurse "$ProjectRoot/PSKoans" -Destination (Join-Path -Path ([System.IO.Path]::GetDirectoryName($Profile)) -ChildPath Modules)`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,7 +42,7 @@ Install-Module Pester -Scope CurrentUser
 1. `git clone` the repository into your desired directory, or download the repository as a .zip file and extract into a directory of your choice.
 2. Then from a normal powershell session run `Get-ChildItem -Recurse | Unblock-File` in that directory to remove the "downloaded from internet" flag that blocks them from running.
 3. Check `Get-ExecutionPolicy`: if it says 'Restricted' or 'Undefined', you need to also run `Set-ExecutionPolicy RemoteSigned` in order to allow the scripts to run.
-4. Run `Copy-Item -Recurse "$ProjectRoot/PSKoans" -Destination ($env:PSModulePath -split ';')[0]`
+4. Run `Copy-Item -Recurse "$ProjectRoot/PSKoans" -Destination (Join-Path -Path ([System.IO.Path]::GetDirectoryName($Profile)) -ChildPath Modules)`
```
vexx32: I would probably prefer still using $env:PSModulePath here in case they decide that has to be changed at some point. I think instead of splitting on `;` you can split on `[System.IO.Path]::PathSeparator` and the rest ought to work correctly?

Tyrix: With the PSModulePath environment variable there is still a big assumption made. What if the first element is not the one pointing to the user documents folder (e.g. on windows)?

Look at this post for example: http://wahlnetwork.com/2015/08/10/psmodulepath/  
What does he have set as the first path? 😉 Who knows which software does that too.

vexx32: Indeed. Rather than that approach, then, given that this is the "dev" setup anyway (because most users should generally install from the gallery) perhaps the recommendation here should simply be to prepend the repo path to the PSModulePath?

```powershell
$env:PSModulePath = "$(Get-Location);$env:PSModulePath"
```

vexx32:
```suggestion
4. Run `$env:PSModulePath = "$(Get-Location);$env:PSModulePath"` from the main repository folder to add it to the PSModulePath and allow you to run the module normally.
```

vexx32:
Changed this in #143.

Thanks for bringing this one up & the suggestions! 😊

PR #112, opened 2018-12-18
Change addition array example - fixes #111

### LINKED ISSUE #111: AboutAssignmentAndArithmetic Arrays Example
[AboutAssignmentAndArithmetic L75](https://github.com/vexx32/PSKoans/blob/master/PSKoans/Koans/Foundations/AboutAssignmentAndArithmetic.Koans.ps1#L75)

```powershell
[int[]]@(1, 2, 3, 4, 5) + '17' | Should -Be @(1, 2, 3, 4, 5, 17)
```

I'm not sure this example teaches the intended lesson. By appending `'17'` to the end of the array declaration like that, PowerShell just writes the `1..5` out, and then appends `'17'` to the output. It does the same thing with strings that can't be coerced to an integer, although the comment on the line before indicates that the value is being converted.

`[int[]]@(1, 2, 3, 4, 5) + '17' | gm` returns member information for both integers and strings, indicating both types of values are written to the pipeline.

At first I thought this was occurring because no type coercion was being performed because no assignment operation was being done. Turns out, nope. Try this example.

```powershell
$a = [int[]]@(1, 2, 3, 4, 5)
$a += '17'
$a += "I love PSKoans"
$a

#outputs
1
2
3
4
5
17
I love PSKoans
```

Version info
```powershell
$psversiontable

PSVersion                      5.1.18300.1001
BuildVersion                   10.0.18300.1001
```

thomasrayner:


PSKoans/Koans/Foundations/AboutAssignmentAndArithmetic.Koans.ps1
```diff
@@ -72,7 +72,7 @@ Describe 'Arithmetic Operators' {
             12.21 + 'FILL_ME_IN' -eq 23.43 | Should -BeTrue
 
             # Adding items into typed arrays will also cause the resulting value to be converted
-            [int[]]@(1, 2, 3, 4, 5) + '17' | Should -Be __
+            [int[]]( @(1, 2, 3, 4, 5) + '17' ) | Should -Be __
```
vexx32: We should use your initial idea from #111 instead and store it in a typed variable to illustrate the original point I was trying to make more clearly, I think.
```suggestion
             [int[]] $Array = @(1, 2, 3, 4, 5)
             $Array += '17' 
             $Array | Should -Be __
```

thomasrayner:
I like that better, too. I think it's more readable for people newer to PowerShell.

PR #114, opened 2018-12-19
Remove unneeded wait periods

vexx32:
* Removes `Start-Sleep` periods from `Show-MeditationPrompt`
* Removes small waiting period from `Invoke-Koan` so that it can return as soon as it detects the threaded job's completion.

/cc @Markekraus @thomasrayner


steviecoaster:
Get those checks working so you can merge it :)

vexx32:
@thomasrayner it helps somewhat. If you want to speed it up any further than this, you probably can, but check into `Invoke-Koan.ps1` instead; that's likely where the biggest gains and losses are to be had. 😄

PR #115, opened 2018-12-20
Reformat Koans to report better error messages

vexx32:
- Should error messages changed to list expected values in a more sensible fashion


vexx32:
<3

PR #118, opened 2018-12-23
Display additional progress bar for farthest topic reached

### LINKED ISSUE #117: Perhaps chapter/milemarkers for better sense of progression.
#Background
I just started on the Koans last night, mostly enjoying it.
Currently, however, shifts from one subject to another are abrupt. "Oh, I guess I'm doing assignment operators now?"

# Describe "Functionality"

Progress is currently shown purely as a "x of (total)". I'm contemplating the idea of having the various subjects being titled and represented by "Mile Markers" in the progress display. Perhaps have them represented by a glyph/symbol, and only display the current and next subject in order to keep the output manageable/readable.

Thoughts?

### LINKED ISSUE #119: Progress Bar throws error under PowerShell ISE environment

When running `Measure-Karma` under ISE the following error is printed in lieu of the progress bar:

```code
Specified argument was out of the range of valid values.
Parameter name: times
At C:\Program Files\WindowsPowerShell\Modules\pskoans\0.42.2\Private\Write-MeditationPrompt.ps1:138 char:20
+                 "$([char]0x2015)" * ($ProgressWidth - $PortionDone)
+                    ~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (:) [], ArgumentOutOfRangeException
    + FullyQualifiedErrorId : System.ArgumentOutOfRangeException
```

## Context "The Problem"

Running `Measure-Karma` in the ISE integrated console triggers the issue consistently.

The progress bar should be printed as per normal

---

This appears to be due to the ISE console not populating the `$host.UI.RawUI.WindowSize` field. It appears it *does* correctly populate the `$host.UI.RawUI.BufferSize` field instead, so the width from that field should also be checked.

vexx32:
image

Resolves #117 
Fixes #119

### DIFF
--- a/PSKoans/Private/Show-MeditationPrompt.ps1
+++ b/PSKoans/Private/Show-MeditationPrompt.ps1
@@ -55,9 +55,32 @@
         [switch]
         $Complete,
 
+        [Parameter(Mandatory, ParameterSetName = 'Meditation')]
+        [ValidateScript(
+            {
+                $MissingKeys = switch ($_) {
+                    { $_.Keys -notcontains 'Name' } { 'Name' }
+                    { $_.Keys -notcontains 'Completed' } { 'Completed' }
+                    { $_.Keys -notcontains 'Total' } { 'Total' }
+                }
+
+                $ErrorString = if ($MissingKeys) {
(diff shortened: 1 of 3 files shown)


vexx32:
@steviecoaster whatcha think? 😄

PR #122, opened 2019-01-04
Load Dummy types (incl. [__]) globally

vexx32:
This is necessary for Pester's -BeOfType to see them.

We can load them conditionally that a type of the same name isn't already loaded with a simple switch, but they need to be in the global scope for Pester's commands to be able to see them.

`using module` won't allow these commands to see the types borrowed from PSKoans in the current scope, it appears to only check types available in Pester's module scope.


vexx32:
Confirmed that problematic file `AboutVariables.Koans.ps1` now is assessed correctly and the types are properly recognised.

PR #126, opened 2019-01-10
Refactor Data Strings & Show-MeditationPrompt

vexx32:
New design:
image

Also moved _all_ data strings to PSD1 data file for cleanliness.


vexx32:
Failing build is due to issue with BuildHelpers module used for CI. Looks like @RamblingCookieMonster has been tinkering around, and for whatever reason one of the build variables isn't able to be bound to a string despite confirming it's not null. One of these sections seems to be the culprit (Error talks about parameter binding exception for a `-Value` parameter).

https://github.com/RamblingCookieMonster/BuildHelpers/blob/fc2d8409f3285fa790b1666dbcfd4619dd107603/BuildHelpers/Public/Set-BuildEnvironment.ps1#L118-L124

https://github.com/RamblingCookieMonster/BuildHelpers/blob/fc2d8409f3285fa790b1666dbcfd4619dd107603/BuildHelpers/Public/Set-BuildEnvironment.ps1#L134

https://github.com/RamblingCookieMonster/BuildHelpers/blob/fc2d8409f3285fa790b1666dbcfd4619dd107603/BuildHelpers/Public/Set-BuildEnvironment.ps1#L135-L137

thomasrayner:
@vexx32 Have you considered manipulating the background color in addition to the foreground color? Some of your choices might look a little wonky for folks using different themes (like some of the high contrast ones).

vexx32:
@thomasrayner I've considered it, but taking into account the fact that those colors may also be modified from the defaults arbitrarily, I'm not sure what I can do in terms of making sure it remains readable?

SeeminglyScience:
@vexx32 If `PSReadLine` is loaded you can pull from it's settings to determine what colors to use.

```powershell
$psrlModule = Get-Module PSReadLine
if (-not $psrlModule) {
    return
}

$options = Get-PSReadLineOption
if ($psrlModule.Version.Major -ge 2) {
    # Handle as ANSI escape
    return '{1}Colored like a number{0} {2}Colored like a keyword{0}' -f (
        "$([char]27)[0m", # Reset color ANSI escape
        $options.NumberColor,
        $options.KeywordColor)
} else {
    # Handle like ConsoleColor
}
```

vexx32:
That... is complicated. But I appreciate the pointer! I'll copy it to an issue so I can keep track of it, but it probably won't happen anytime soon 😕

thomasrayner:
Maybe in the meantime, it would be easier to implement a `Set-PSKoansColorScheme -Mode <fancy | consoledefault>` for people to opt in and out

vexx32:
🤔 

Yeah, I can see why you'd need that, I suppose. I'll think on it. 😄

thomasrayner:
> 🤔
> 
> Yeah, I can see why you'd need that, I suppose. I'll think on it. 😄

Just to eliminate all the colors in general and use whatever bland colors the console is set to. Maybe colorblind people have a setting they like and are annoyed by the fanciness 🤷‍♂️

vexx32:
See #127. Yep, the option will be there... sooner or later 😉

PR #135, opened 2019-02-06
Adding Blank Class to fix #131

### LINKED ISSUE #131: Issue with the AboutArrays.Koans.ps1 file

I noticed that the last test 'does not throw exceptions with invalid indexes' in the Foundations\AboutArrays.Koans.ps1 file passes without any user modifications.

https://github.com/vexx32/PSKoans/blob/622f2739bc4d70596ffe522876dc1735d7d148e3/PSKoans/Koans/Foundations/AboutArrays.Koans.ps1#L94-L103

Both sides of the Pipeline are $null.  The left side has the alias named __, and that calls the function Get-Blank that returns $null.  On the right side the Should -Be values are outside of the range of the array, and that also returns $null.

## Context "Your Attempts"

I had attempted to change the __ to 0, which now fails and the following is displayed when running `Measure-Karma`.

```
The answers you seek...

Expected $null, but got 0.

    Please meditate on the following code:

[It] does not throw exceptions with invalid indexes
at <ScriptBlock>, C:\Users\Username\PSKoans\Foundations\AboutArrays.Koans.ps1: line 99
99:         0 | Should -Be $Array[4]
```

I'm running the following versions:
PSversion: 5.1.17763.134
pskoans: 0.42.2
pester: 4.6.0

tysonjhayes:
This was bugging me and @vexx32 made it super easy to correct.


vexx32:
@tysonjhayes looks like we're running into some scoping issues when it comes to where the type is and is not defined.

Best approach might just be to move this one to the external types .PS1 file. The existing classes there can be replaced by or can just inherit this type.

tysonjhayes:
Sorry I'm not sure where you are referring to when you say "external types." I don't see anything called that (I might be missing something totally obvious).

vexx32:
Sorry, I'd forgotten what I called them. They're here:

https://github.com/vexx32/PSKoans/blob/master/PSKoans/DummyTypes.ps1

tysonjhayes:
I wasn't able to get the pester tests to pass adding the class into the dummytypes.ps1 file. So I added it's own class/folder/file which is how I've side stepped this in the past. If you don't want me to do that I'm happy to try something else.

tysonjhayes:
Apparently I just fail at this, these are passing on my machine before I submit them. 😬

PSKoans/Classes/Blank.ps1
```diff
@@ -0,0 +1,9 @@
+class Blank {
+    [string] ToString() {
+        return $null
+    }
+    
+    [bool] op_Equals([object] $other) {
+        return $false
+    }
+}
```
vexx32: Pretty sure this will need to be in the dummy types file so that it's externally visible.

PSKoans/Public/Get-Blank.ps1
```diff
@@ -20,11 +20,11 @@
         https://github.com/vexx32/PSKoans
     #>
     [CmdletBinding()]
-    [OutputType([void])]
+    [OutputType([Blank])]
```
vexx32: Reference the type name as a string rather than the type object itself; it'll defer looking the type up until the function is called, by which time the type should definitely be in scope.
```suggestion
    [OutputType('Blank')]
```

PSKoans/Public/Get-Blank.ps1
```diff
@@ -20,11 +20,11 @@
...
     [Alias('__', 'FILL_ME_IN')]
     param()
 
     Write-Verbose "I AIN'T DOIN' NOTHIN'!!!"
 
-    $null
+    return New-Object -TypeName:Blank
```
vexx32: This function is called a _lot_ in the koans, and the additional overhead incurred with `New-Object` will be a bit of a drag on the overall performance when evaluating,
```suggestion
    [Blank]::new()
```

Tests/Functions/Public/Get-Blank.Tests.ps1
```diff
@@ -3,6 +3,7 @@
 Describe 'Get-Blank' {
 
     It 'should not produce output' {
+        $breakvar = $true;
```
vexx32: This doesn't seem to be used. Is it needed anywhere? 🙂

PSKoans/Public/Get-Blank.ps1
tysonjhayes: Totally fair, was not aware of the performance overheard of New-Object, I'll start call ::new() from now on. Thanks for the tip!

And because I trust but verify I ran some numbers.

```powershell
Measure-Command  {
>>     for ($i = 0; $i -lt 10000; $i++) {
>>         [Blank]::New()
>>     }
>> }


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 76
Ticks             : 768778
TotalDays         : 8.89789351851852E-07
TotalHours        : 2.13549444444444E-05
TotalMinutes      : 0.00128129666666667
TotalSeconds      : 0.0768778
TotalMilliseconds : 76.8778


Measure-Command {
>>     for ($i = 0; $i -lt 10000; $i++) {
>>         New-Object -TypeName:Blank
>>     }
>> }


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 1
Milliseconds      : 173
Ticks             : 11733668
TotalDays         : 1.35806342592593E-05
TotalHours        : 0.000325935222222222
TotalMinutes      : 0.0195561133333333
TotalSeconds      : 1.1733668
TotalMilliseconds : 1173.3668


```

Tests/Functions/Public/Get-Blank.Tests.ps1
tysonjhayes: Yup, thought I removed that. It's what I use for setting breakpoints. I'll get that removed.

PSKoans/Classes/Blank.ps1
tysonjhayes: Doesn't seem like it, I just updated the output type to encapsulate as a string vs calling the object and that seems to have corrected the issue as its now passing the tests here.

PSKoans/Public/Get-Blank.ps1
vexx32: 👍 wise policy!

In general, language features will always be faster than entire cmdlets; if there's a language feature that does it, use that instead. 😄

PSKoans/Classes/Blank.ps1
vexx32: Interesting! Well, as long as it works, I'm happy with it. 😄

vexx32: Last thing (I promise!)

Please make sure there's a newline at the end of the file here. 🙂

tysonjhayes: Usually my VS code does this for me, must have lost the setting somewhere. Fixed.

PR #136, opened 2019-02-06
Provider Clearer Context around Write-Output vs Return

### LINKED ISSUE #134: Provider Clearer Context around Write-Output vs Return

AboutFunctionsAndScriptBlocks.Koans.ps1

Koan could use clearer context around Write-Output vs return statements.

The following code could be clearer on what is happening specifically around that Write-Output can return multiple values to the pipe where return stops all output and doesn't run any code after it.

### Original Code
```powershell
 It 'allows you to assign a name to a sequence of commands' {
        <#
            An apt name eases the transition to abstraction.
            PowerShell function names are named in convention following
            the cmdlets; Verb-Noun. To see the list of approved
            PowerShell verbs, run 'Get-Verb' in the console.

            There are many methods of sending output from a function.
        #>
        function Get-Number {
            <#
                Explicit use of Write-Output allows multiple items to be output
                within the function without causing the function to exit.
            #>
            Write-Output 20
        }
        function Get-ReturnedNumber {
            <#
                Use of the return statement causes all output previously declared
                to be sent, along with any output passed to the return statement
                itself, then ends the function.
            #>
            return 13
        }
        function Get-DroppedNumber {
            <#
                Leaving a value, or a statement that returns a value on a line
                by itself causes that output to be 'dropped' to the output
                stream, similar to functional languages. It can be considered
                an implicit (and faster) 'Write-Output'
            #>
            12
        }

        20 | Should -Be (Get-Number)
        Get-ReturnedNumber | Should -Be 13
        12 | Should -Be (Get-DroppedNumber)
    }
```

### Modified Code

```powershell
    It 'allows you to assign a name to a sequence of commands' {
        <#
            An apt name eases the transition to abstraction.
            PowerShell function names are named in convention following
            the cmdlets; Verb-Noun. To see the list of approved
            PowerShell verbs, run 'Get-Verb' in the console.

            There are many methods of sending output from a function.
        #>
        function Get-Number {
            <#
                Explicit use of Write-Output allows multiple items to be output
                within the function without causing the function to exit.
            #>
            Write-Output 20
            Write-Output 42
        }
        function Get-ReturnedNumber {
            <#
                Use of the return statement causes all output previously declared
                to be sent, along with any output passed to the return statement
                itself, then ends the function.
            #>
            return 13
            Write-Output "Doe not return"
        }
        function Get-DroppedNumber {
            <#
                Leaving a value, or a statement that returns a value on a line
                by itself causes that output to be 'dropped' to the output
                stream, similar to functional languages. It can be considered
                an implicit (and faster) 'Write-Output'
            #>
            12
        }

        20,42 | Should -Be (Get-Number)
        Get-ReturnedNumber | Should -Be 13
        12 | Should -Be (Get-DroppedNumber)
    }
```

Happy to provide a pull request if there is interest.

tysonjhayes:
Fixes #134

### DIFF
--- a/PSKoans/Koans/Foundations/AboutFunctionsAndScriptBlocks.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutFunctionsAndScriptBlocks.Koans.ps1
@@ -49,7 +49,7 @@ Describe 'Functions' {
             12
         }
 
-        __ __ | Should -Be (Get-Number)
+        __ | Should -Be (Get-Number)
         Get-ReturnedNumber | Should -Be 13
         __ | Should -Be (Get-DroppedNumber)
     }
--- a/PSKoans/Koans/Foundations/AboutFunctionsAndScriptBlocks.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutFunctionsAndScriptBlocks.Koans.ps1
@@ -28,6 +28,7 @@ Describe 'Functions' {
                 within the function without causing the function to exit.
             #>
             Write-Output 20
+            Write-Output 42
         }
         function Get-ReturnedNumber {
             <#
@@ -36,6 +37,7 @@ Describe 'Functions' {
                 itself, then ends the function.
             #>
             return __
+            Write-Output "Does not return anything."
         }
         function Get-DroppedNumber {
             <#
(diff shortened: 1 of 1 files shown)


PSKoans/Koans/Foundations/AboutFunctionsAndScriptBlocks.Koans.ps1
```diff
@@ -47,7 +49,7 @@ Describe 'Functions' {
             12
         }
 
-        __ | Should -Be (Get-Number)
+        __ __ | Should -Be (Get-Number)
```
vexx32: Not sure what happened here, but `Get-Blank` doesn't take any arguments. 🙂

PR #138, opened 2019-02-08
Update AboutStrings to work cross-platform

### LINKED ISSUE #137: AboutStrings is not cross-platform
The `Evaluated Strings` context in `Foundations/AboutStrings.Koans.ps1` calls `Get-Item 'C:\Windows'` which will always fail on non-Windows machine.

`Evaluated Strings` context:

```PowerShell
        It 'can do a simple expansion' {
            $Windows = Get-Item 'C:\Windows' | Select-Object -ExpandProperty FullName
            '__' | Should -Be "The windows directory is located here: $Windows"
        }
```

## Context "Your Attempts"
There are few ways this could be fixed.  If you prefer to use `Get-Item` without a variable,  you could use something like `Get-Item '.'`.  Or if you don't mind variables you could use something like `Get-Item $env:HOME`.  Or you could something less platform specific like

```PowerShell
        It 'can do a simple expansion' {
            $DayOfWeek = Get-Date | Select-Object -ExpandProperty DayOfWeek
            '__' | Should -Be "The day of the week today is $DayofWeek"
        }
```

puthuppu:
Fixes #137

Since the idea is to just show simple variable expansion I thought it would be easiest to just use `$pwd` instead of creating a new variable.

### DIFF
--- a/PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
@@ -52,7 +52,7 @@ Describe 'Strings' {
 
         It 'handles other ways of doing the same thing' {
             # Strings can handle entire subexpressions being inserted as well!
-            $String = "Your default profile exists?: $(Test-Path $PROFILE)"
+            $String = "Your home folder is: $(Get-Item $HOME)"
             '__' | Should -Be $String
         }
 
--- a/PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
@@ -47,8 +47,7 @@ Describe 'Strings' {
         }
 
         It 'can do a simple expansion' {
-            $Windows = Get-Item 'C:\Windows' | Select-Object -ExpandProperty FullName
-            '__' | Should -Be "The windows directory is located here: $Windows"
+            '__' | Should -Be "Your working directory is located here: $pwd"
         }
 
         It 'handles other ways of doing the same thing' {
--- a/PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
@@ -47,12 +47,12 @@ Describe 'Strings' {
         }
 
         It 'can do a simple expansion' {
-            '__' | Should -Be "Your working directory is located here: $pwd"
+            '__' | Should -Be "Your home directory is located here: $HOME"
         }
 
         It 'handles other ways of doing the same thing' {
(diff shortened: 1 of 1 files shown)


PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
```diff
@@ -47,8 +47,7 @@ Describe 'Strings' {
         }
 
         It 'can do a simple expansion' {
-            $Windows = Get-Item 'C:\Windows' | Select-Object -ExpandProperty FullName
-            '__' | Should -Be "The windows directory is located here: $Windows"
+            '__' | Should -Be "Your working directory is located here: $pwd"
```
vexx32: $pwd isn't a bad choice, but it is one that will cause problems if the end user tries to run `Measure-Karma` from another directory. Additionally, there isn't actually a guarantee that the koans will be evaluated with the same current working directory; the koans are evaluated in subprocesses that are completely removed from the user's own session.

Using a more consistent variable such as $home may be more preferable, or $env:TEMP

PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
```diff
@@ -47,13 +47,12 @@ Describe 'Strings' {
...
         It 'can do a simple expansion' {
-            $Windows = Get-Item 'C:\Windows' | Select-Object -ExpandProperty FullName
-            '__' | Should -Be "The windows directory is located here: $Windows"
+            '__' | Should -Be "Your home directory is located here: $HOME"
         }
 
         It 'handles other ways of doing the same thing' {
             # Strings can handle entire subexpressions being inserted as well!
-            $String = "The windows directory is located at $(Get-Item 'C:\Windows')"
+            $String = "Your default profile exists?: $(Test-Path $PROFILE)"
```
vexx32:
```suggestion
            $String = "Your PowerShell profile path is '$PROFILE'"
```

If only because many people new to PowerShell may not be aware of profiles, do this koan, and then later create one... and suddenly this koan would fail again. 😄

puthuppu: Does that demonstrate "entire subexpressions" being inserted as well?  Or isn't that essentially the same as the previous example of "simple expansion"?  How about something like:

```PowerShell
"You are running the $(Get-Module -Name PSKoans) module!"
```

vexx32: Hmm, you're right. That would probably work. However, if the user has more than one module of that name it will come back with `System.Object[]` 😄 

Perhaps something more like the original, just cross-platform?
```suggestion
            $String = "Your default profile exists?: $(Get-Item $HOME)"
```

PR #141, opened 2019-03-18
Update AboutComparison.Koans.ps1

JonWalz2:
-in is not the inverse of -contains

### DIFF
--- a/PSKoans/Koans/Foundations/AboutComparison.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutComparison.Koans.ps1
@@ -126,7 +126,7 @@ Describe 'Comparison Operators' {
 
     Context '-in and -notin' {
 
-        It 'is the inverse of -contains' {
+        It 'is equivalent to -contains' {
             $Array = 1, 2, 3, 4, 5
             $SearchValue = __
 


vexx32:
🤔 Getting into tricky semantic territory here, no? You're right in that it _performs the same function_, but in its **usage** it's completely the opposite with regards to which side of the operator you put the array and the scalar on.

I'm not against changing it at all, but I think saying "they're the same" sends the wrong message here. Do you have a more accurate alternative, perhaps? 🙂

JonWalz2:
It is tricky semantic territory. When I saw the assertion that: "It is the inverse of -contains" I was expecting to see a test that used -notcontains which would be the inverse of -contains. Now I understand that what you meant is that the order is reversed. Perhaps an assertion such as, "It finds a value in an array" would be clearer?

vexx32:
I think that would be best, yeah absolutely. 🙂

PSKoans/Koans/Foundations/AboutComparison.Koans.ps1
```diff
@@ -126,7 +126,7 @@ Describe 'Comparison Operators' {
 
     Context '-in and -notin' {
 
-        It 'is the inverse of -contains' {
+        It 'is equivalent to -contains' {
```
vexx32:
```suggestion
        It 'finds a value in an array' {
```

vexx32:
I'm adding a variation of this in #143.

It's been a while since I got a release out, so I'm trying to get everything tidied up so we can get a new version out.

Thanks for pointing this out! 😄

PR #142, opened 2019-03-27
Make Clear-Host optional for Measure-Karma

### LINKED ISSUE #139: Remove Clear-Host from Measure-Karma function
# Describe "Functionality"

Please remove the Clear-Host command from the Measure-Karma function

## Context "Private or Public? What are the use cases? Parameters? Options?"

If someone want's a clear screen they can enter the Clear-Host command before Measure-Karma. I prefer to be able to control when my screen is cleared. This could even be a switch parameter if you want to remove the Measure-Karma text from the screen when running the command.

vexx32:
Fixes #139 

* Adds `-ClearScreen` switch to `Measure-Karma` to trigger `Clear-Host`.

Users can set `$PSDefaultParameterValues['Measure-Karma:ClearScreen'] = $true` if they would like this to be their default.

### DIFF
--- a/PSKoans/Public/Measure-Karma.ps1
+++ b/PSKoans/Public/Measure-Karma.ps1
@@ -68,7 +68,12 @@
 
         [Parameter(Mandatory, ParameterSetName = "Reset")]
         [switch]
-        $Reset
+        $Reset,
+
+        [Parameter()]
+        [Alias()]
+        [switch]
+        $ClearScreen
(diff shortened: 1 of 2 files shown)


vexx32:
@steviecoaster @stevenjudd @shaneis if you fellas could give this a quick once over and confirm for me that it does what it oughtta, would be much appreciated! 😄

PR #189, opened 2019-05-26
Clarify AboutPSProviders->FileSystemProvider Koan.

### LINKED ISSUE #174: Issues with 'Filesystem Provider'

https://github.com/vexx32/PSKoans/blob/e0a54b0363c84ef67262a34dad48caa9fd745e02/PSKoans/Koans/Cmdlets%201/AboutPSProviders.Koans.ps1#L130

### Context 'Filesystem Provider'

I was a bit confused about this whole context. To begin with, I think that it needs more explanations, especially clarifying that `TestDrive:` is a drive created by PSKoans. 

---
https://github.com/vexx32/PSKoans/blob/e0a54b0363c84ef67262a34dad48caa9fd745e02/PSKoans/Koans/Cmdlets%201/AboutPSProviders.Koans.ps1#L149
Since we insert random newlines in the tempfile, its `Length` is random on every `Measure-Karma`, so answer would be `$File.Length | Should -Be $File.Length`

If the file is created like this, it would be `10000` every time, and we would have newlines in the file.
```
-join (1..10000 | ForEach-Object {
    Get-Random -Minimum 32 -Maximum 96 |
    ForEach-Object {
        if ($_ -match '[25]0$'){
            "`n"
        } else {
            [char]$_ -as [string]
        }
    }
}) | Set-Content -Path $Path
```
---
https://github.com/vexx32/PSKoans/blob/e0a54b0363c84ef67262a34dad48caa9fd745e02/PSKoans/Koans/Cmdlets%201/AboutPSProviders.Koans.ps1#L154
This would also be random every time? Is the answer `$FirstLine | Should -Be $FirstLine`?

BrandonLundt:
﻿# PR Summary

Resolves #174

Instead of using the TempDrive created by Pester, added logic to determine execution environment and use the local temp directory.

Instead of dynamically creating a text file, replaced with static lines of text. Makes the following tests clearer.
## Changes

- Replaced dynamically generated file with static text.

## Checklist

- [x] Summarised changes.

### DIFF
--- a/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
@@ -127,15 +127,16 @@ Describe 'Environment Provider' {
 }
 
 Describe 'FileSystem Provider' {
-    $Path = 'TestDrive:' | Join-Path -ChildPath 'File001.tmp'
+    BeforeAll {
+        $Path = 'TestDrive:' | Join-Path -ChildPath 'File001.tmp'
 
-    $FileContent = @'
+        $FileContent = @'
 PSKOANS!
 The Env: drive contains system environment data. Its contents can vary wildly from OS to OS,
 especially between Windows, Mac, and Linux, for example.
 '@
-    Set-Content -Path $Path -Value $FileContent
-
+        Set-Content -Path $Path -Value $FileContent
+    }
     It 'allows access to various files and their properties' {
         $File = Get-Item -Path $Path
 
--- a/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
@@ -128,17 +128,13 @@ Describe 'Environment Provider' {
 
 Describe 'FileSystem Provider' {
     $Path = 'TestDrive:' | Join-Path -ChildPath 'File001.tmp'
-    if (-not (Test-Path $Path)) {
-        New-Item -Path $Path > $null
 
-        -join (1..10000 | ForEach-Object {
-                Get-Random -Minimum 32 -Maximum 96 |
-                    ForEach-Object {[char]$_ -as [string]}
-                if ((Get-Random) -match '[25]0$') {
-                    "`n"
-                }
-            }) | Set-Content -Path $Path
-    }
+    $FileContent = @'
+PSKOANS!
+The Env: drive contains system environment data. Its contents can vary wildly from OS to OS,
+especially between Windows, Mac, and Linux, for example.
+'@
+    Set-Content -Path $Path -Value $FileContent
 
     It 'allows access to various files and their properties' {
         $File = Get-Item -Path $Path


PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
```diff
@@ -127,17 +127,33 @@ Describe 'Environment Provider' {
...
+            $Path = Join-Path -Path \tmp -ChildPath 'PSKoans.tmp'
+        }
+        else{
+            #Path to MacOS temp directory
+            $Path = Join-Path -Path $env:TMPDIR -ChildPath 'PSKoans.tmp'
+        }
+    }
+    else{
+        Write-Error -Message "Was not able to determine temp directory location."
+    }
```
vexx32: Couple things here.

If `\tmp` is meant to be for Linux/Mac FS, you need to use forward slashes; their filesystem won't work with backslashes. 

Also, this is too complicated, especially for something we're setting up in a koan topic file. If we really need this boilerplate, I'd prefer it gets put in the PSM1 as a script-scoped variable and access to that data provided via a public module function (e.g., `Get-TempDirectory` or some such).

Relatedly, I don't want to be writing non-Pester-related errors from a koan lesson file. All it's going to do is confuse the user, unfortunately.

PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
```diff
@@ -127,17 +127,33 @@ Describe 'Environment Provider' {
...
-                if ((Get-Random) -match '[25]0$') {
-                    "`n"
-                }
-            }) | Set-Content -Path $Path
+
+        "PSKOANS!
+        The Env: drive contains system environment data. Its contents can vary wildly from OS to OS,
+        especially between Windows, Mac, and Linux, for example.
+
+        The only shared Env: items across all OS's currently are Path and PSModulePath." | Add-Content -Path $Path
```
vexx32: This might perhaps be better formatted as a here string, I think?

BrandonLundt: I will change this to a here string. I think it is important to keep this snippet as part of the koan, so when reading through the test it is clear what content is in the file. Otherwise, one would have to browse to their temp directory and open a file. Doesn't seem helpful when trying to complete:
https://github.com/vexx32/PSKoans/blob/f18528243e64ccc4006844ca2088e2ee8f7a3e62/PSKoans/Koans/Cmdlets%201/AboutPSProviders.Koans.ps1#L167-L169

vexx32: Agreed.

PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
BrandonLundt: 1. Done, nice catch!
2. I started this off as a private function, and suppressed any errors. Let me know if you would like further changes.

PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
```diff
@@ -127,17 +127,19 @@ Describe 'Environment Provider' {
 }
 
 Describe 'FileSystem Provider' {
-    $Path = 'TestDrive:' | Join-Path -ChildPath 'File001.tmp'
+    #Get-TempDirectory is a private function in PSKoans
+    #This should return a path to your temp directory.
+    $Path = Get-TempDirectory
```
vexx32: Private functions aren't visible to the koans during execution. It'll need to be a public function, unfortunately. 🤔 Pester has TestDrive, I wonder if they expose any functions for this... Hm, doesn't look like it unfortunately.  Except for `TestDrive:\` -- might be a simpler alternative given it's an already-available solution, perhaps?

I'm not sure.

/cc @nohwnd is `TestDrive:` reliable on unix systems?

PSKoans/Private/Get-TempDirectory.ps1
```diff
@@ -0,0 +1,36 @@
...
+        Finds the temp directory on the current OS.
+
+    .DESCRIPTION
+        Finds the temp directory on the current OS, appends the filename PSKoans.tmp and returns a string of the path.
+
+    .EXAMPLE
+        Get-TempDirectory
+
+        Returns C:\Users\PSKoans\AppData\Local\Temp\PSKoans.tmp
+    #>
```
vexx32: Looks good, apart from the mentioned complication with private functions...

If this is going to be a public function, we'll need to generate markdown help for this with PlatyPS so that it creates the proper external help MAML for this function as well. You could do this or I can take care of it as a follow up. 🙂 

Assuming we stick with this course, at least. `TestDrive:` might be a simpler way to go, depending...

PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
```diff
@@ -127,17 +127,19 @@ Describe 'Environment Provider' {
...
-            }) | Set-Content -Path $Path
+
+        $FileContent = @'
+PSKOANS!
+The Env: drive contains system environment data. Its contents can vary wildly from OS to OS,
+especially between Windows, Mac, and Linux, for example.
+
+The only shared Env: items across all OSs currently are Path and PSModulePath.
+'@
+        Set-Content -Path $Path -Value $FileContent
```
vexx32: This section is for the FileSystem provider, though. The text itself might seem a little out of place in context, I'm thinking?

PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
nohwnd: Yes :) it should work at least as well as on windows.

PSKoans/Private/Get-TempDirectory.ps1
BrandonLundt: Agreed, TestDrive seems the simpler approach. In rereading #174 the issue is less about using TestDrive and more about not knowing the contents of the file. The addition of the here string *should* resolve that confusion. 

So I would vote we get rid of get-tempdirectory and leverage testdrive.

vexx32: Sounds good to me. 😄

vexx32:
🤔 Hmm, looks like you pulled in some commits with a merge or something? You might need to `rebase -i master` or something and trim those out.

nohwnd:
Switch to this branch. Then do `git branch "file-system-provider-backup"`. This will checkout another branch from this branch to make sure you don't lose the commits. Then do `git reset head~3` to reset the last 3 commits, then `git stash -u` and `git stash drop` to throw away the changes coming from the last three commits. if you now do `git log` you should see the commits first two commits but not the last three. once you are happy `git push --force-with-lease` to update the branch on the remote, which will also update this merge request. (or you can do rebase but that would mean that you'd need to solve merge conflicts if there were any, so this approach seems easier :D)

vexx32:
Yeah but then he'd rose also his actual last commit. Rebase is simpler in my mind... Eh, whatever works. :P

nohwnd:
Ah okay, there we commits to keep. In that case interactive rebase is indeed simpler.

BrandonLundt:
@vexx32  Yeah, instead of doing git pull like i should have to update my local branch to the changes of the project master, I did git fetch then git rebase FETCH_HEAD thinking I would be able to skip a merge commit. Long story short, just being a dummy. At this point, I *think* the issue related to this Pull Request is actually resolved AND git history is coherent. Let me know if you want anything done differently.

@nohwnd  I haven't used git push --force-with-lease before / didn't know it existed. Thanks for the advanced! Worked well.

PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
```diff
@@ -128,17 +128,13 @@ Describe 'Environment Provider' {
...
-                }
-            }) | Set-Content -Path $Path
-    }
+
+    $FileContent = @'
+PSKOANS!
+The Env: drive contains system environment data. Its contents can vary wildly from OS to OS,
+especially between Windows, Mac, and Linux, for example.
+'@
+    Set-Content -Path $Path -Value $FileContent
```
vexx32: We should probably have this in a `BeforeAll {}` block, I think?

BrandonLundt: Makes sense to me, I have updated accordingly.

PR #192, opened 2019-05-27
Fix divide by zero on some systems

### LINKED ISSUE #178: Attempted to divide by zero on Show-MeditationPrompt.ps1
When I use "Measure-Karma -Topic" and the name of any completed topic after it, I get an "Attempted to divide by zero."  message in the output. Although this is not a serious bug but I thought it worth mentioning.

This is the complete error message:

Attempted to divide by zero.
At C:\Program Files\WindowsPowerShell\Modules\PSKoans\0.50.1\Private\Show-MeditationPrompt.ps1:172 char:17
+ ...             $PortionDone = ($KoansPassed / $TotalKoans) * $ProgressWi ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], RuntimeException
    + FullyQualifiedErrorId : RuntimeException

My PowerShell and module versions are:

Name           Version
----                -------
PSKoans        0.50.1

PSVersion                      5.1.17763.503
BuildVersion                   10.0.17763.503

tomlarse:
Proposed fix #178 

﻿# PR Summary

Use `$TotalKoans` in stead of `$PesterTests.TotalCount` in both conditions. 

## Checklist

- ['x'] Pull Request has a meaningful title.
- ['x'] Summarised changes.
- ['x'] Pull Request is ready to merge & is not WIP.
- ['x'] Added tests / only testable interactively.
- ['NA'] Added documentation / opened issue to track adding documentation at a later date.

### DIFF
--- a/Tests/Functions/Public/Measure-Karma.Tests.ps1
+++ b/Tests/Functions/Public/Measure-Karma.Tests.ps1
@@ -132,7 +132,7 @@ Describe 'Measure-Karma' {
 
             It 'should not divide by zero if all Koans are completed' {
                 $KoansCompletedTestLocation = 'TestDrive:{0}PSKoansCompletedTest' -f [System.IO.Path]::DirectorySeparatorChar
-                $TestFile = Join-Path -Path $TestLocation -ChildPath 'SingleTopicTest.Koans.Ps1'
+                $TestFile = Join-Path -Path $KoansCompletedTestLocation -ChildPath 'SingleTopicTest.Koans.Ps1'
 
                 New-Item $KoansCompletedTestLocation -ItemType Directory
                 New-Item $TestFile -ItemType File
--- a/Tests/Functions/Public/Measure-Karma.Tests.ps1
+++ b/Tests/Functions/Public/Measure-Karma.Tests.ps1
@@ -131,10 +131,10 @@ Describe 'Measure-Karma' {
             }
 
             It 'should not divide by zero if all Koans are completed' {
-                $TestLocation = 'TestDrive:{0}PSKoansCompletedTest' -f [System.IO.Path]::DirectorySeparatorChar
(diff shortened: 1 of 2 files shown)


vexx32:
🤔 We really need a Pester test for this functionality to make sure we don't hit this again.

Think you can add a Pester test to the module tests that does the following?

1. `Set-PSKoanLocation` to `TestDrive:\TestKoans\` (and create the folder)
2. Adds a basic koan file with only one test, that is already passing. The text for this file can be embedded in the test itself with a here-string or whatever seems appropriate. Named something like `SingleTopicTest.Koans.ps1` is fine.
3. Runs `Measure-Karma -Topic` against that file by name and verifies that it doesn't throw.

tomlarse:
Something like this?

Tests/Functions/Public/Measure-Karma.Tests.ps1
```diff
@@ -129,6 +129,30 @@ Describe 'Measure-Karma' {
...
+param()
+
+Describe 'Koans Test' {
+    It 'is easy to solve' {
+        $true | should -be $true
+    }
+}
+'@ | Set-Content $TestFile
+
+                Set-PSKoanLocation $TestLocation
```
vexx32: We should probably retrieve the current location first, so we can set it back to whatever it was after this, just to minimise potential future impact on other tests.

Tests/Functions/Public/Measure-Karma.Tests.ps1
```diff
@@ -129,6 +129,30 @@ Describe 'Measure-Karma' {
...
+
+                {Measure-Karma -Topic SingleTopicTest} | should -Not -Throw
```
vexx32:
```suggestion
                { Measure-Karma -Topic SingleTopicTest } | Should -Not -Throw
```

PR #198, opened 2019-06-06
Changed out-file parameter and updated Out-String koan

### LINKED ISSUE #197: AboutOutCmdlets
Line 84: Out-File parameter must be FilePath instead of Path

❌ 'Stored knowledge is of little value until it is used.' | Out-File -Path $Path -NoNewline
✔️ 'Stored knowledge is of little value until it is used.' | Out-File -FilePath $Path -NoNewline

Line 125: Apparently there is a carriage return character (`r) in $String variable in addition to every New Line character (`n) in each line which is not the case with here-string so the test will not pass even if I copy and paste the exact output of $String.

         Expected: '\r\n\r\n    Di...'
         But was:  '\n\n        ...'

BrandonLundt:
﻿# PR Summary
Resolves #197 

## Context

Out-String Koan is changed in its approach as carriage returns are "funny" and potentially confusing when trying to compare Get-Item output in a here string. The idea is if the here string is static, then the user only has to update what is piping into out-string. Since this koan is about out-string this feels appropriate.
## Changes

- Changed Out-File parameter from -Path to -FilePath to support Windows PowerShell 5.1 and earlier.
- Changed Out-String koan to not require here-string edits, instead requires updates to what is piped into Out-String.

## Checklist

- [x] Summarised changes.

### DIFF
--- a/PSKoans/Koans/Cmdlets 2/AboutOutCmdlets.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 2/AboutOutCmdlets.Koans.ps1
@@ -115,7 +115,7 @@ Context 'Out-* Cmdlets' {
             Out-String is a fantastic and simple way to turn complex objects into strings that look just
             like PowerShell's default console formatting.
         #>
-    
+
         It 'creates string representations of data' {
             #Create a hashtable that pipes into Out-String
             $String = @{ } | Out-String
--- a/PSKoans/Koans/Cmdlets 2/AboutOutCmdlets.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 2/AboutOutCmdlets.Koans.ps1
@@ -81,7 +81,7 @@ Context 'Out-* Cmdlets' {
         It 'stores data in files' {
             $Path = 'TestDrive:\File.txt'
 
-            'Stored knowledge is of little value until it is used.' | Out-File -Path $Path -NoNewline
+            'Stored knowledge is of little value until it is used.' | Out-File -FilePath $Path -NoNewline
 
             '__' | Should -Be (Get-Content -Path $Path)
         }
@@ -115,23 +115,20 @@ Context 'Out-* Cmdlets' {
             Out-String is a fantastic and simple way to turn complex objects into strings that look just
             like PowerShell's default console formatting.
         #>
-
+    
         It 'creates string representations of data' {
-            $HomeFolder = Get-Item -Path $HOME
-
-            $HomeFolder | Should -Not -BeNullOrEmpty
-            $String = $HomeFolder | Out-String
-
-            # Fill in this here-string with the data you'd see in your console from calling Get-Item $HOME
+            #Create a hashtable that pipes into Out-String
+            $String = '__' | Out-String
+    
             @"
-
-
-    Directory: __
-
-
-Mode                LastWriteTime         Length Name
-----                -------------         ------ ----
-______         __________________                ____
+    
+    Name                           Value                                                                                                                                                                                         
+    ----                           -----                                                                                                                                                                                         
+    ColumnA                        Value1                                                                                                                                                                                        
+    ColumnB                        Value2                                                                                                                                                                                        
+    
+    
+    
 "@ | Should -Be $String
             # Mind the indentations; here-strings have to terminate at the very beginning of a line.
         }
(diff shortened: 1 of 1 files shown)


PSKoans/Koans/Cmdlets 2/AboutOutCmdlets.Koans.ps1
```diff
@@ -115,23 +115,20 @@ Context 'Out-* Cmdlets' {
...
+    
         It 'creates string representations of data' {
-            $HomeFolder = Get-Item -Path $HOME
-
-            $HomeFolder | Should -Not -BeNullOrEmpty
-            $String = $HomeFolder | Out-String
-
-            # Fill in this here-string with the data you'd see in your console from calling Get-Item $HOME
+            #Create a hashtable that pipes into Out-String
+            $String = '__' | Out-String
```
vexx32: If we're asking for a hashtable we should probably provide at minimum an empty hashtable or just an unquoted blank I guess as by this point I guess it's reasonable to expect them to know how to do that at this point.

PSKoans/Koans/Cmdlets 2/AboutOutCmdlets.Koans.ps1
```diff
@@ -115,23 +115,20 @@ Context 'Out-* Cmdlets' {
...
-
-
-Mode                LastWriteTime         Length Name
-----                -------------         ------ ----
-______         __________________                ____
+    
+    Name                           Value                                                                                                                                                                                         
+    ----                           -----                                                                                                                                                                                         
+    ColumnA                        Value1                                                                                                                                                                                        
+    ColumnB                        Value2
```
vexx32: Given we're using a hashtable here (_fantastic_ idea btw, tons simpler than my original idea on both sides of the fence), we should probably swap the `Column` names for `Key` names to better reflect what's expected here. 🙂

PSKoans/Koans/Cmdlets 2/AboutOutCmdlets.Koans.ps1
```diff
@@ -115,23 +115,20 @@ Context 'Out-* Cmdlets' {
...
-----                -------------         ------ ----
-______         __________________                ____
+    
+    Name                           Value                                                                                                                                                                                         
+    ----                           -----                                                                                                                                                                                         
+    ColumnA                        Value1                                                                                                                                                                                        
+    ColumnB                        Value2                                                                                                                                                                                        
+    
+    
+
```
vexx32: I think there's one too many newlines here compared to what will typically be returned from Out-String; this will probably be enough to fail the test.

BrandonLundt: Looks like VSCode helped me out with some formatting. :)

PSKoans/Koans/Cmdlets 2/AboutOutCmdlets.Koans.ps1
BrandonLundt: Works for me. Updated.

PSKoans/Koans/Cmdlets 2/AboutOutCmdlets.Koans.ps1
BrandonLundt: Agreed.

PSKoans/Koans/Cmdlets 2/AboutOutCmdlets.Koans.ps1
```diff
@@ -115,23 +115,20 @@ Context 'Out-* Cmdlets' {
             Out-String is a fantastic and simple way to turn complex objects into strings that look just
             like PowerShell's default console formatting.
         #>
-
+
```
vexx32: Looks like we added some whitespace on this line. Best to remove it just for cleanliness' sake. 🙂

BrandonLundt: Done.

PSKoans/Koans/Constructs and Patterns/AboutSplatting.Koans.ps1
```diff
@@ -86,4 +86,4 @@ Describe 'Splatting' {
             (Get-ChildItem @Parameters).Count | Should -Be 2
         }
     }
-}
+}
```
vexx32: Last little thing!

Files should have a trailing newline for cross-platform reasons (aka Linux is weird in how it defines things but hey). 😄

PR #200, opened 2019-06-08
Removed -Recurse, updated Regex Pattern, added help

### LINKED ISSUE #193: Issues with AboutSplatting
I'm using this example to demonstrate the problem. For some reason that I couldn't find these two lines are not equal :

                   $x1 = Get-ChildItem C:\Test
                   $x2 = Get-ChildItem C:\Test

running $x1 | Should -Be $x2 will return this error:
            
                  Expected @(Test1-1, Test1-2, Test1-3), but got @(Test1-1, Test1-2, Test1-3).
                  At C:\Program Files\WindowsPowerShell\Modules\Pester\4.8.1\Functions\Assertions\Should.ps1:186 char:9
                  +         throw ( New-ShouldErrorRecord -Message $testResult.FailureMes ...
                  +                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                  + CategoryInfo          : InvalidResult: (System.Collections.Hashtable:Hashtable) [], 
                   Exception
                  + FullyQualifiedErrorId : PesterAssertionFailed

so no matter what format we use for Get-ChildItem parameters (LongLines, Escaping, Hashtable) on lines 28-43, the test will not pass.

I'm not sure if comparing $x1 and $x2 using ($x1 -eq $x2) is right or not, but if it is, it won't return true or false :thinking: and using ($x1 -ne $x2) will print out one of them I think, no true or false again. Maybe this is by design and the reason why the tests don't pass. :thinking: 

One more thing: I think using -Recurse with -Depth doesn't seem to have a meaning, since using -Depth will override the effect of -Recurse. So these will return the same thing:

                   $x1 = Get-ChildItem C:\Test -Depth 2
                   $x2 = Get-ChildItem C:\Test -Recurse -Depth 2

$env:PSKoans:KoanFolder is null:
I'm not sure if this variable populates when running Measure-Karma or not, but running it by itself won't return anything and I think the only reason this command returns the PSKoans folder's content is that I opened up the PSKoans folder in VScode and my current path in terminal is PSKoans folder, so running Get-ChildItem -Path $env:PSKoans:KoanFolder will return the current folder despite -path parameter being null. So I replaced it with $HOME+"\PSKoans" for now.

There is no file in PSKoans root folder (Only folders) so this test will not pass
                    
                  Get-ChildItem @Parameters | Select-Object -First 1 | Should -BeOfType System.IO.FileInfo

It's better to change line 60 to something like this or add the depth parameter in the if statement:

                $Parameters = @{
                       Path = $HOME+"\PSKoans"
                       Depth = 2
                }

I couldn't figure out the last part (Wildcard), I think a little bit more explanation would be helpful.

BrandonLundt:
﻿# PR Summary

Resolves #193 
## Context

Complex objects like Get-ChildItem will always fail the ` | should -be ` comparison. Comparing two properties of that object seems sufficient to prove that they are in fact the same after using the splatting operator. If the dual comparison is too much, could easily reduce to comparing just .name

 The last It block was inheritently difficult because the regex pattern no longer matches to any of the directory names. https://github.com/vexx32/PSKoans/blob/1edc1b91e6f92eae19dec46bb019c9222477218d/PSKoans/Koans/Constructs%20and%20Patterns/AboutSplatting.Koans.ps1#L80

Minor changes to the regex pattern and some additional comments should make it clear how to pass this Koan.
## Changes

- Remove -Recurse parameter from Get-ChildItem calls. There are plenty of examples later on that show how to include a switch parameter in a splat.
- Updated regex to include digits and '_'
- Added comments surrounding -match and $Matches

## Checklist

- [x] Summarised changes.

### DIFF
--- a/PSKoans/Koans/Constructs and Patterns/AboutSplatting.Koans.ps1
+++ b/PSKoans/Koans/Constructs and Patterns/AboutSplatting.Koans.ps1
@@ -28,20 +28,18 @@ Describe 'Splatting' {
             # Here are a few common ways a detailed command to find files in a folder might be written:
 
             # 1. Super long lines; hard to follow along with.
-            $LongLines = Get-ChildItem -Path $PSKoansFolder -Include '*.ps1' -Recurse -Depth 2
+            $LongLines = Get-ChildItem -Path $PSKoansFolder -Include '*.ps1' -Depth 2
 
             # 2. Escaping linebreaks; more readable, but very fragile and prone to errors
             $Escaping = Get-ChildItem `
                 -Path $PSKoansFolder `
                 -Include '*.ps1' `
-                -Recurse `
                 -Depth 2
 
             # 3. Splatting using a hashtable. Note the similarity to #2 and fill in missing values.
             $Parameters = @{
                 Path    = $PSKoansFolder
                 Include = '__'
-                Recurse = $true # Switches can be assigned a boolean value.
                 Depth   = __
             }
             $Splatted = Get-ChildItem @Parameters
@@ -49,10 +47,14 @@ Describe 'Splatting' {
                 All above approaches are equal in effect, but splatting is a much tidier and more
                 maintainable approach.
             #>
-            $Splatted | Should -Be $Escaping
-            $Escaping | Should -Be $LongLines
+            $Splatted.Count | Should -Be $Escaping.Count
+            $Splatted.Name | Should -Be $Escaping.Name
 
-            $LongLines | Should -Be $Splatted
+            $Escaping.Count | Should -Be $LongLines.Count
+            $Escaping.Name | Should -Be $LongLines.Name
+
+            $LongLines.Count | Should -Be $Splatted.Count
+            $LongLines.Name | Should -Be $Splatted.Name
         }
 
     }
@@ -65,6 +67,7 @@ Describe 'Splatting' {
             $Value = __
             if ($Value -eq 7) {
                 $Parameters.Add('File', $true)
+                $Parameters.Add('Recurse', $true)
             }
             else {
                 $Parameters.Add('Directory', $true)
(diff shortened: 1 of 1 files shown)


PSKoans/Koans/Constructs and Patterns/AboutSplatting.Koans.ps1
```diff
@@ -65,6 +67,7 @@ Describe 'Splatting' {
             $Value = __
             if ($Value -eq 7) {
                 $Parameters.Add('File', $true)
+                $Parameters.Add('Recurse', $true)
```
vexx32: Why add the -Recurse here?

PSKoans/Koans/Constructs and Patterns/AboutSplatting.Koans.ps1
```diff
@@ -77,13 +80,22 @@ Describe 'Splatting' {
...
+                If $String = "Folders:00_TheBasics"
+                Then $Matches holds the following:
+                Name                           Value
+                ----                           -----
+                Filter                         00_TheBasics
+                0                              Folders:00_TheBasics
+                #>
                 $Matches.Remove(0) # Remove the 'whole' match and keep only the portions we asked for
-                $Matches.Clone()
+                $Matches.Clone() #Matches.Clone() allows for the remaining matches to be assigned to $Parameters
```
vexx32:
```suggestion
                # Clone() copies the remaining matches, so we can store them with the $Parameters assignment above.
                $Matches.Clone()
```

PSKoans/Koans/Constructs and Patterns/AboutSplatting.Koans.ps1
```diff
@@ -77,13 +80,22 @@ Describe 'Splatting' {
...
             }
             $Parameters.Add('Path', $PSKoansFolder)
             $Parameters.Add('Directory', $true)
-            (Get-ChildItem @Parameters).Count | Should -Be 2
+            #Now w
+            (Get-ChildItem @Parameters).Count | Should -Be __
```
vexx32: Partially missing comment? 🙂

PSKoans/Koans/Constructs and Patterns/AboutSplatting.Koans.ps1
BrandonLundt: Because there are no files at the root of $PSKoansFolder.

PSKoans/Koans/Constructs and Patterns/AboutSplatting.Koans.ps1
BrandonLundt: Yeah, literally no idea what comment I was going to leave here. So I have cleaned up that line.

PSKoans/Koans/Constructs and Patterns/AboutSplatting.Koans.ps1
```diff
@@ -77,13 +80,23 @@ Describe 'Splatting' {
...
-            $Parameters = if ($String -match 'Folders:(?<Filter>[a-z ])$') {
+            $Parameters = if ($String -match 'Folders:(?<Filter>[_\d\w ]*)$') {
+                <#Matches is automaticed populated by the -match operater in the if statement
+                If $String = "Folders:00_TheBasics"
+                Then $Matches holds the following:
+                Name                           Value
+                ----                           -----
+                Filter                         00_TheBasics
+                0                              Folders:00_TheBasics
+                #>
```
vexx32: Can we reformat this to match the block-comment style used elsewhere in the koans? Usually I keep the opening/closing lines clear and indent the comments, like so:

```powershell
<#
    Comments here.
#>
```

Issue #212, opened 2019-07-01
AutoCompleted "It"'s in AboutComparison.Koans.ps1


EVWorth:
# Describe "Koan Bug, Issue, or Help Request"
There are three sections that are auto-completing progress even though the progress is not completed.
Once reaching AboutComparison, Running Measure-Karma displays 3/19 complete before beginning.

https://github.com/vexx32/PSKoans/blob/3070afd4200c997572ce5950bc91d44a1014d26f/PSKoans/Koans/Foundations/AboutComparison.Koans.ps1#L19-L26

https://github.com/vexx32/PSKoans/blob/3070afd4200c997572ce5950bc91d44a1014d26f/PSKoans/Koans/Foundations/AboutComparison.Koans.ps1#L169-L173

https://github.com/vexx32/PSKoans/blob/3070afd4200c997572ce5950bc91d44a1014d26f/PSKoans/Koans/Foundations/AboutComparison.Koans.ps1#L192-L197

all three of these show as completed/green when running .\AboutComparison.Koans.ps1

## Context "The Problematic Assertions"

for instance, with: 
https://github.com/vexx32/PSKoans/blob/3070afd4200c997572ce5950bc91d44a1014d26f/PSKoans/Koans/Foundations/AboutComparison.Koans.ps1#L169-L173

If I remove the first $true, and replace it with __, the Koan shows the it statement failing.

## Context "Your Attempts"

the following are the options and their success/failure for the 3 first "words" for "return $true if either input is $true'

$true, __, __ : Pass
$true, $true, $true : Pass
__ , $true, $true : Fail
$true, $true, $false : Fail

## Context "Additional Information"

Pester Version: 4.8.1
PowerShell Version: 5.1.18362.145

vexx32:
Ah, gotcha. Okay, yeah, these need to be changed. The assertion `__ | Should -BeTrue` (or anything equivalent) will always pass. I think a lot of these could/should be fixed by replacing `__` entries with `$____` -- the variable doesn't exist so it will fail the test properly. 🙂

EVWorth:
would it make sense to replace all instances of __ with $__ and '__' with '$__' in all Koans?

vexx32:
No, not in all instances. At a glance, it looks appropriate here because we're expecting the variable `$true` or `$false` to be input (therefore a blank variable `$____` makes sense).

There are a lot more places we're expecting numbers, text, etc. -- the format of the blank should generally match the expected input. 🙂

Issue #213, opened 2019-07-01
Autocompleted 'It' in AboutStrings.Koans.ps1


EVWorth:
# Describe "Koan Bug, Issue, or Help Request"

After Finishing Koan 107 (Comparison), I ran Measure-Karma and already showed 1/18 finished for AboutStrings.

![image](https://user-images.githubusercontent.com/50814635/60465360-87055d00-9c16-11e9-8afa-f909b7057589.png)

## Context "The Problematic Assertions"

I ran .\AboutStrings.Koans.ps1 and saw the following:

![image](https://user-images.githubusercontent.com/50814635/60465745-7a353900-9c17-11e9-9be3-3cf9effdd99a.png)

Looking at the file, it looks like the Koan is missing its request.
![image](https://user-images.githubusercontent.com/50814635/60465863-b8caf380-9c17-11e9-86ee-2b852a1ab9c1.png)

' ' seems to be the correct answer, but it looks like the Koan should have '__' so the user can remove it.

vexx32:
I agree, this shouldn't complete automatically. I think just replacing one of the `$String` values with a blank is a good way to go. The literal space can stay as is, maybe?

vexx32:
Thanks for closing these, I quite forgot!

@EVWorth for future reference when you do a PR, you can actually have related issues closed automatically when a PR is merged if you mention something along the lines of `Resolves #ISSUENUMBER` in the PR summary. That lets Github know to automatically close the issue when the PR is merged in.

There's a [short list of keywords](https://help.github.com/en/articles/closing-issues-using-keywords) Github recognises for this, so you can use what you prefer. 🙂

EVWorth:
No problem!

I thought there was something like that.

![image](https://user-images.githubusercontent.com/50814635/60593624-c9d74a00-9d68-11e9-8a40-1f788c35d7ac.png)
Looks like thats what I get for being too wordy!

vexx32:
Hahaha so close.... 😄

PR #214, opened 2019-07-02
Fixes auto-succeeding pester tests in AboutComparison and AboutStrings

EVWorth:
﻿# PR Summary

This PR fixes pester tests from auto-succeeding.

This PR Resolves Issues #212 and #213 

I tested these changes to make sure they initially fail the pester tests successfully

## Context

The solution for the AboutComparison issues came from: https://github.com/vexx32/PSKoans/issues/212#issuecomment-507712857

The solution for the AboutString issues came from: https://github.com/vexx32/PSKoans/issues/213#issuecomment-507713959

## Changes

In AboutStrings:
- I replaced text in $String1 on line 105 that was causing the test to succeed without user input. 

In AboutComparison.Koans.ps1
-  I replaced 4 instances of __ with $__, where the pester tests were succeeding without user input, because __ was evaluating to $true which completed the test. 

## Checklist

- [X] Pull Request has a meaningful title.
- [X] Summarised changes.
- [X] Pull Request is ready to merge & is not WIP.
- [X] Added tests / only testable interactively.
- [X] Added documentation / opened issue to track adding documentation at a later date.

### DIFF
--- a/PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
@@ -102,7 +102,7 @@ Describe 'Strings' {
 
         It 'adds strings together' {
             # Two become one.
-            $String1 = 'This string'
+            $String1 = '__'
             $String2 = 'is cool.'
 
             $String1 + ' ' + $String2 | Should -Be 'This string is cool.'
--- a/PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
@@ -102,7 +102,7 @@ Describe 'Strings' {
 
         It 'adds strings together' {
             # Two become one.
-            $String1 = '__'
+            $String1 = '_____'
             $String2 = 'is cool.'
 
             $String1 + ' ' + $String2 | Should -Be 'This string is cool.'
--- a/PSKoans/Koans/Foundations/AboutComparison.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutComparison.Koans.ps1
@@ -22,7 +22,7 @@ Describe 'Comparison Operators' {
 
         It 'is a simple test' {
             $true -eq $false | Should -Be $false
-            __ | Should -Be (1 -eq 1)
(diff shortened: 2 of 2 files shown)


vexx32:
Looks good! Can we use $____ rather than $__ (four underscores instead of two) just so it's very clear we're not trying to reference $_?

EVWorth:
Good point/idea!

Done!

PSKoans/Koans/Foundations/AboutStrings.Koans.ps1
```diff
@@ -102,7 +102,7 @@ Describe 'Strings' {
 
         It 'adds strings together' {
             # Two become one.
-            $String1 = 'This string'
+            $String1 = '__'
```
vexx32:
```suggestion
            $String1 = '_____'
```

Been trying to move towards using long-blanks for strings as well, just to make it more clear that we're expecting more than a couple of characters. 🙂

EVWorth:
I like the longer blanks. They stand out more which is always a good thing. Do you have a preference on standardizing the number of underscores?

perhaps 4 of each? ____ for int/double, `"____"` for string, and $____ for Boolean? 

I know that the Foundations Koans could use a good freshening up in that regard. sounds like a project...  😄

vexx32:
@EVWorth yeah it's changed over time as I realize this might be a bit clearer for various reasons. I'd say.... if we're expecting: 
* Numeric input, go with `__` 
* Strings, go with `'____'`
* Variables, go with `$____`
* Command names or strings in command arguments, go with `____` unless the string you expect would normally require quotes to enter it at a command line.

Of course, with strings it's good to think about whether the string you expect might make use of interpolation and should thus use double quotes instead of single quotes. 🙂

EVWorth:
Ok awesome. I'll use that going forward. I was thinking about the ' ' vs "" thing earlier. I think it makes sense to differentiate between those as well.

vexx32:
🤔 Sigh, I need to add a contributing style guide lmao

EVWorth:
If you do I'll contribute to it and ask you 50 billion questions you don't want to have to answer ;)

vexx32:
I don't mind answering haha, it's all good. Just something I should do. I guess I'll make an issue so I don't forget. :)

PR #274, opened 2019-10-01
Check for PSKoan directory before opening

### LINKED ISSUE #256: Show-Karma -Meditate doesn't check the folder exists

Executing `Show-Karma -Meditate` before the first run does not first check to see if the koans folder actually exists first.

## Context "The Problem"

There is no check in the code for whether the folder exists.

https://github.com/vexx32/PSKoans/blob/da66f062c93dca112214016285b2bd5891eb87c5/PSKoans/Public/Show-Karma.ps1#L45-L66

The command should behave similarly to the standard `Show-Karma` behaviour, where it checks for the folder's existence. If it does not exist, it should execute `Update-PSKoan` the same way that the default (parameterless) behaviour does.

```powershell
Name    Version
----    -------
PSKoans 0.50.1
PSKoans 0.50.0
PSKoans 0.42.2

PSVersion                      5.1.17763.592
BuildVersion                   10.0.17763.592
```

usrme:
# PR Summary

Fixes #256 

This simple `if` statement should (as far as my own testing went) fix the issue where the existence of the folder isn't checked prior to it being opened.

## Context

The goal of this pull request is to make sure the existence of the PSKoans directory is always checked, so a situation where the directory has been removed, but is being acted upon will not occur.

## Changes

All changes localized to *PSKoans/Public/Show-Karma.ps1*:

- `Write-Verbose` statements for additional information
- `if` statement using `Test-Path`
- `Update-PSKoan` usage as per the default parameterless behaviour

## Checklist

* [x]  Pull Request has a meaningful title.
* [x]  Summarised changes.
* [x]  Pull Request is ready to merge & is not WIP.
* [ ]  Added tests / only testable interactively.
  * Make sure you add a new test if old tests do not effectively test the code changed.
* [ ]  Added documentation / opened issue to track adding documentation at a later date.


usrme:
Sure thing. I'll get on that as soon as possible!

usrme:
I'm awfully new to Pester and I'm trying to wrap my head around the testing procedure at the moment; hopefully you can show me the light a little. As far as I am understanding within the 'With -Contemplate Switch' I need to add two sets of additional contexts: one for when the actual directory exists (not just the TestDrive one?) and one for when the directory doesn't exist?

Any additional hints you can give are much appreciated!

vexx32:
Hmm... I would copy that whole Context block and make a new one, with an appropriate name (`With -Meditate switch and folder doesn't exist` or something along those lines) and copy as much of it makes sense in that context, remove anything that isn't needed. :slightly_smiling_face:

usrme:
Alright, these officially constitute as the very first Pester tests I've ever written. Feedback and additional resources for educating myself are more than appreciated!

Tests/Functions/Public/Show-Karma.Tests.ps1
```diff
@@ -217,5 +217,28 @@ Describe 'Show-Karma' {
...
+
+        Context 'With -Meditate Switch' {
+
+            Context 'PSKoans directory exists' {
+                BeforeAll {
+                    Mock Get-PSKoanLocation { "$HOME/PSKoans" }
+                }
+
+                It 'should find the PSKoan directory' {
+                    Test-Path -Path (Get-PSKoanLocation) | Should -Be $true
```
vexx32: Ah, I see where the misunderstanding is. Np, np. 🙂 

OK, so what you need to do is mock `Test-Path` , `Invoke-Item`, and `Get-Command` above, and then test the `Show-Karma -Meditate` switch by calling that command directly. I think one of the existing tests already has a section for the `-Meditate` switch, which you can literally just move down to this section under the `directory exists` context. Then all you need to do is copy that logic and adjust it for the case where the directory doesn't exist. You'd need to mock `Update-PSKoan` for that case as well, since all we need to verify is that that command gets called.

After all, the purpose of the test is to show that `Show-Karma` is behaving correctly, which we can't be sure of until we actually call the command. 🙂 

If any of that isn't clear just let me know, I hope I'm not leaving you to fumble about in the dark too much! 😊

usrme:
I'm just desperately trying to get my mind to wrap around what is going on behind the scenes with these mocks and assertions! I even bought Adam Bertram's 'The Pester Book' to try and make sense of it all, but it hasn't clicked for me yet thus far.

As far as I am understanding:

- I need to mock `Test-Path` to `$false` because the `if (-not ...)` statement I added needs to evaluate to `$true`
- I need to mock `Update-PSKoan` because I don't want to actually write anything to the user's file system while testing
- I need to mock `Get-Command` to `$false` because I want to bypass all the following `if` statements in `Show-Karma.ps1` until the `else` block
- I need to mock `Invoke-Item` because I don't want anything to actually open on the user's computer
- run `Show-Karma -Meditate` in the `It` block because I want to actually test whether all the relevant commands (mocks) get called
  - as verified by the `Assert-MockCalled` cmdlets

I removed the `PSKoans directory exists` context (maybe erroneously) because to my limited understanding it's already being implicitly checked within the previous `With -Contemplate Switch` context.

Now, I am having a hard time with actually checking that the PSKoans directory gets created as a result of the test. I tried doing `Test-Path (Get-PSKoanLocation) | Should -Be $true` right below `Show-Karma -Meditate` in the `It` block, but that never evaluated to `$true`, so I tried mocking `Get-PSKoanLocation` as such, but that didn't help either:

```
Mock Get-PSKoanLocation { Join-Path $TestDrive 'PSKoans' }
```

I am more than happy to pick this "problem" up in Slack or wherever else if this is too spammy for the pull request as it's becoming apparent to me that I need some mentoring with this to help the process of understanding along.

vexx32:
To answer your question @usrme, you don't need to verify that the folder actually gets created. That's what the `Update-PSKoan` command would do, so that is covered by the tests for that command. Mocking it and assuring it's called properly is all we need to do for this command. 🙂 

And since you've mocked it, it won't do anything regardless; once you `Mock` a command you're basically completely overwriting it so that you can just ensure that the commands that are being called are the ones you expect to be called.

And yeah, you're right, the case where the directory _does_ already exist is indeed already covered. We could potentially restructure tests a bit so they're laid out a bit better, but that's a pretty minor thing and can be handled another day. 

But yeah, good work, I hope it wasn't too frustrating for you! 😊 💖

PR #276, opened 2019-10-01
 Fix "Show-Advice" requiring full basename for snippet

### LINKED ISSUE #261: Show-Advice -Name requires the full basename

## Context "The Problem"

Unlike with `Show-Karma -Topic <name>`, `Show-Advice -Name <name>`  requires the topic name to include the `.Advice` suffic (`Show-Karma` doesn't expect the `.Koans` suffix to be included).

It also does not support tab completion the same way that Show-Karma's -Topic parameter does.

`Get-Advice -Name Profile` should show the `Profile.Advice.json` snippet.

usrme:
# PR Summary

Fixes #261 

## Context

Fixed 'Show-Advice' cmdlet requiring full basename to print advice snippet and improved cmdlet's support for tab completion, which was non-existent before. 

There are no tests for this change, but I am more than willing to try and implement them, but before I tack something on: should those tests be created into a separate _Show-Advice.Tests.ps1_ tests file or is there already an existing set of tests for advices that I am not seeing?

## Changes

- Hardcoded _Advice._ to `Get-ChildItem` filter to _Show-Advice.ps1_
  - If the format of the advice snippet is standardized to always be _Advice.json_, then is hardcoding still an issue?
- Added `Show-Advice` and `Get-Advice` to array of commands in _RegisterArgumentCompleters.ps1_
- Created separate 'completer' region for **Name** parameter

## Checklist

- [x] Summarised changes.

### DIFF
--- a/PSKoans/Init/RegisterArgumentCompleters.ps1
+++ b/PSKoans/Init/RegisterArgumentCompleters.ps1
@@ -4,6 +4,8 @@ $CommandName = @(
     'Reset-PSKoan'
     'Show-Karma'
     'Update-PSKoan'
+    'Show-Advice'
+    'Get-Advice'
 )
 
 #region Topic completer
@@ -42,6 +44,31 @@ $RegisterParams = @{
 Register-ArgumentCompleter @RegisterParams
 
 $RegisterParams.ParameterName = 'IncludeModule'
+Register-ArgumentCompleter @RegisterParams
+
+#endregion
+
+#region Name completer
+
+$RegisterParams = @{
+    CommandName   = $CommandName
+    ParameterName = 'Name'
+    ScriptBlock   = {
+        param($Command, $Parameter, $WordToComplete, $CommandAst, $FakeBoundParams)
+
+        $Values = Get-Module PSKoans |
+            Select-Object -ExpandProperty ModuleBase |
+            Join-Path -ChildPath 'Data/Advice' |
+            Get-ChildItem -File -Recurse |
+            Select-Object @{
+                Name       = 'Name';
+                Expression = { $_.BaseName.Replace('.Advice', '') }
+            } |
+            Select-Object -ExpandProperty Name
+
+        return @($Values) -like "$WordToComplete*"
+    }
+}
 Register-ArgumentCompleter @RegisterParams
 
 #endregion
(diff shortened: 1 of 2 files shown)


PSKoans/Init/RegisterArgumentCompleters.ps1
```diff
@@ -45,3 +47,28 @@ $RegisterParams.ParameterName = 'IncludeModule'
...
+#region Name completer
+
+$RegisterParams = @{
+    CommandName   = $CommandName
+    ParameterName = 'Name'
+    ScriptBlock   = {
+        param($Command, $Parameter, $WordToComplete, $CommandAst, $FakeBoundParams)
+
+        $Values = Get-Module PSKoans |
+            Select-Object -ExpandProperty ModuleBase |
```
vexx32: Not a bad way to go, but you can target `$script:ModuleRoot` to get this path as this file is dot-sourced into the psm1. :slightly_smiling_face:

usrme: I'll change this tomorrow!

vexx32:
@usrme there aren't currently tests for the *-Advice commands. I don't expect you to add tests for everything (though you're free to if you'd like!) but tests for this should be in a new file, yeah. :sparkling_heart:

usrme:
I changed the other completer's pipeline to `$script:ModuleRoot` as well for consistency. I won't be adding tests just yet as I feel I still have a lot to learn about testing with Pester.

PSKoans/Init/RegisterArgumentCompleters.ps1
```diff
@@ -4,6 +4,8 @@ $CommandName = @(
     'Reset-PSKoan'
     'Show-Karma'
     'Update-PSKoan'
+    'Show-Advice'
+    'Get-Advice'
```
vexx32: We should put the `*-Advice` command names in a different variable or redefine `$CommandName` immediately before registering the argumentcompleter below, so we don't accidentally register completers for commands they're not valid for. 🙂

vexx32:
I've opened #278 to track missing tests for these commands. We'll get them written sooner or later, it's not an especially high priority at the present. 🙂

PSKoans/Init/RegisterArgumentCompleters.ps1
usrme: Ah, yes, that makes perfect sense. I'll get on that as soon as I can.

PR #285, opened 2019-10-11
:bug: Fix fun advices filenames

robdy:
﻿# PR Summary
I noticed that the cmdlets `Show-Advice -Name Fun*` are not working and gives the following error:
image

It's caused by wrong naming (missing `.Advice` part) so I renamed them.

## Context
No context, just my observation while playing with the tool

## Changes
- File renamed `FunWithGUIs.json` =>`FunWithGUIs.Advice.json`
- File renamed `FunWithCats.json` =>`FunWithCats.Advice.json`

## Checklist

- [x] Summarised changes.


vexx32:
Oh, look at that, I completely missed that. Thanks for this! 

Can you also open a new issue so I remember we also need to update those commands to throw a more clear error when it can't find a file? The error message is very confusing and doesn't give proper information. 😁

PR #292, opened 2019-10-20
Update-PSKoan - Fix "Yes to All" for ShouldProcess Prompts

vexx32:
﻿# PR Summary

Currently, Update-PSKoan prompts for every file updated, even if the user selects "A" for "Yes to All" at one of the prompts.

Fix is to have the ShouldProcess check only in the iterating function, and allow the child function to use that preference instead of checking each time it is called.

## Context

The reason this is needed appears to be that "yes to all" is only respected for the duration of the current function. Since `Update-PSKoanFile` was being called once for every file that needed updating, the prompt was unable to remember the user's selection from the prior call. 

Moving the ShouldProcess test into the parent `Update-PSKoan` function appears to resolve the issue most neatly.

## Changes

* Move `$PSCmdlet.ShouldProcess()` test from `Update-PSKoanFile` itself into `Update-PSKoan` and test before calling the private function.

## Checklist

- [x] Summarised changes.

### DIFF
--- a/PSKoans/Private/Update-PSKoanFile.ps1
+++ b/PSKoans/Private/Update-PSKoanFile.ps1
@@ -48,47 +48,45 @@ function Update-PSKoanFile {
             }
         } | Group-Object -Property ID -AsHashTable -AsString
 
-        if (-not $moduleKoans) {
-            # Handles topics which do not have It blocks.
-            return
-        }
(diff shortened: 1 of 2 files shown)


vexx32:
@indented-automation mind giving this one a looksee for me? 💖 😊

PR #289, opened 2019-10-14
Fix bug with VSCode opening empty file

robdy:
﻿# PR Summary
Fixes #284.
## Changes
- Added `$PSKoanLocationFullPath` variable 
- Used the variable for opening VSCode, VSCode for Insiders and folder

## Checklist

- [x] Pull Request has a meaningful title.
- [x] Summarized changes.
- [x] Pull Request is ready to merge & is not WIP.
- [x] Added tests / only testable interactively.
  - Make sure you add a new test if old tests do not effectively test the code changed.
- [ ] Added documentation / opened issue to track adding documentation at a later date.

Issue #284, opened 2019-10-10
Show-Karma -Meditate not opening VSCode properly


robdy:
# Describe "Module Bug or Issue"

I have VSCode installed on my machine. I noticed that when I run `Show-Karma -Meditate`, it doesn't open the folder properly. Instead, it opens `.\~\PSKoans` file (non-existing). You can see that behavior on the screenshot:

![2019-10-10 21_47_52-● PSKoans - Visual Studio Code](https://user-images.githubusercontent.com/15113729/66601331-f2f94900-eba7-11e9-971c-de27257ae7a0.png)

## Context "The Problem"

I cloned the repo directly from GitHub, added the folder to `$env:PSModulePath`, run `Show-Karma` and then, as per the instruction, run `-Meditate`.

The issue sounds to be here:

https://github.com/vexx32/PSKoans/blob/3fb8b44ac9e4d79daec480d397818c4b7a6caa9b/PSKoans/Public/Show-Karma.ps1#L78-L83

In my case, if I change it to this (to use full folder path instead):

```powershell
 $VSCodeSplat = @{ 
     FilePath     = 'code' 
     ArgumentList = '"{0}"' -f (Get-Item (Get-PSKoanLocation)).FullName
     NoNewWindow  = $true 
 } 
 Start-Process @VSCodeSplat 
```
it works correctly:
![image](https://user-images.githubusercontent.com/15113729/66601769-f3461400-eba8-11e9-8e82-ef42614a040e.png)

## Context "Expected Behavior"

I expect the folder to be opened in VSCode.

## Context "Additional Information"

```powershell
PS C:\Users\robdy> Get-Module -Name PSKoans -ListAvailable |
>>     Select-Object -Property Name, Version

Name    Version
----    -------
PSKoans 0.50.1


PS C:\Users\robdy>
PS C:\Users\robdy> $PSVersionTable | Out-String

Name                           Value
----                           -----
PSVersion                      5.1.18362.145
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.18362.145
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

I can open PR to address this, as described above. Alternatively, I can change `Get-PSKoanLocation` to return full path.

vexx32:
🤔 Is this in the module version currently on the gallery?

[...]

PR #308, opened 2019-10-27
Properly quote path argument to Start-Process

### LINKED ISSUE #306: Show-Karma -Meditate doesn't open editor correctly when a space exists in user path

Show-Karma -Meditate doesn't open editor correctly when a space exists in user path

## Context "The Problem"

My user folder contains a space so when `Show-Karma -Meditate` runs it opens a folder path that doesn't exist. `Start-Process` doesn't allow for a space in the path unless it's properly quoted.

The koan folder should open to the correct folder.

```powershell
Show-Karma -Meditate
```

vexx32:
﻿# PR Summary

Start-Process doesn't pass through non-explicit quotes.
This unfortunately means paths with spaces get read as two arguments.

Fix is to properly insert explicit quotes to the argument first.

## Context

There's no reason not to support spaces in paths. This is 2019, c'mon. 😋 

Resolves #306 

## Changes

- Tidy up path resolution logic.
- Refactor Start-Process splat table to insert explicit quotes in the path argument.

## Checklist

- [x] Summarised changes.

### DIFF
--- a/PSKoans/Public/Show-Karma.ps1
+++ b/PSKoans/Public/Show-Karma.ps1
@@ -59,25 +59,28 @@ function Show-Karma {
             Get-PSKoan @GetParams
         }
         'OpenFolder' {
-            $PSKoanLocationFullPath = $pscmdlet.GetUnresolvedProviderPathFromPSPath((Get-PSKoanLocation))
+            $KoanLocation = Get-PSKoanLocation
             Write-Verbose "Checking existence of koans folder"
-            if (-not (Test-Path $PSKoanLocationFullPath)) {
+            if (-not (Test-Path $KoanLocation)) {
                 Write-Verbose "Koans folder does not exist. Initiating full reset..."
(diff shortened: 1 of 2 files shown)


camusicjunkie:
Your modified code looks like what was previously implemented before #289. Will this create the issue again described in #284?

vexx32:
@camusicjunkie the issue in #284 arose because we didn't think to use either Resolve-Path or Get-Item before passing the path to Start-Process. Provided we do that, this should continue to work.

The reason I reverted to a more similar pattern to the original is mainly because I suspect the additional complexity of the methods utilized in #289 made it harder for us to catch the regression here in terms of proper handling of pants with spaces during review.

If you'd like to verify for yourself, you can pull the built module from the CI artifacts and test to be sure it behaves as you'd expect. :blush: 

I'll look into possible avenues of expanding our testing to ensure we don't get tripped up like this again! :sparkling_heart:

PR #317, opened 2019-10-29
Adding -Continue parameter for Show-Karma

arefg:
I think it would be a good idea to have a parameter for Show-Karma cmdlet to be able to not only open the PSKoans directory but also jump to the failing topic and the exact line on that file.

I've added a parameter named OpenFile with an alias (Continue), this parameter uses the $Results.Meditation to extract the path and line number of the failing topic.

Maybe there is a better way of extracting the file path and achieving the same result (by RegEx or other means) but it's working great for VS Code and I think we can change it a little bit to detect the PowerShell ISE and at least open the failing topic file or at best jump to the failing line as well.

With this, we can continue our journey of enlightenment 😄  by just typing:

Show-Karma -Continue


PSKoans/Public/Show-Karma.ps1
```diff
@@ -80,6 +85,30 @@ function Show-Karma {
...
+            try {
+                $Results = Get-Karma @GetParams
+            }
+            catch {
+                $PSCmdlet.ThrowTerminatingError($_)
+            }
+            $UnformatedPath     = $Results.Meditation
+            $PSKoansFilePath = $UnformatedPath.Substring($UnformatedPath.IndexOf(', ')+2,$UnformatedPath.IndexOf(': ')-($UnformatedPath.IndexOf(', ')+2))
+            $PSKoansLine = $UnformatedPath.Substring($UnformatedPath.IndexOf(': line ')+7,$UnformatedPath.LastIndexOf(("`n"))-($UnformatedPath.IndexOf(': line ')+7))    
+            $FullArgument = "-g ${PSKoansFilePath}:${PSKoansLine}"
```
vexx32:
```suggestion
            $FullArgument = '-g "{0}":{1}' -f $PSKoansFilePath, $PSKoansLine
```

We need to have quotes around the path, in case the path has a space in it. I'd also recommend using simpler variable names; `$LineNumber` and `$FilePath` are sufficient here, I think. 🙂

PSKoans/Public/Show-Karma.ps1
```diff
@@ -34,6 +34,11 @@ function Show-Karma {
         [switch]
         $Contemplate,
 
+        [Parameter(Mandatory, ParameterSetName = 'OpenFile')]
+        [Alias('Continue')]
+        [switch]
+        $OpenFile,
```
vexx32: 🤔 Not sold on the name here as yet. I'm also thinking we could just roll this into the `-Contemplate` set as an optional switch, so the syntax would end up being `-Contemplate -File` or something?

PSKoans/Public/Show-Karma.ps1
```diff
@@ -80,6 +85,30 @@ function Show-Karma {
...
+                $EditorSplat = @{
+                    FilePath     = $Editor
+                    ArgumentList = $FullArgument
+                    NoNewWindow  = $true
+                }
+                Start-Process @EditorSplat
+            }
+            else {
+                $PSKoanLocationFullPath | Invoke-Item
+            }
```
vexx32: Somewhere in here we need to check the editor in use, I think? Since these parameters are only valid for VS Code...

If it's not `code` or `code-insiders` then we need to just default to opening the file; we can pave the way forward with a `switch` statement on `$Editor`, which would allow us to easily extend it later with alternate editors and the parameters they support for this kind of thing.

If it _is_ VS Code I'd recommend we also add the `-r` switch so that it opens in the current window and they don't end up with a dozen VS Code windows open. 🙂 

Also, the `else` here should probably just directly open the koans file itself rather than the folder, since the user would have asked for the file itself to be opened.

PSKoans/Public/Show-Karma.ps1
```diff
@@ -80,6 +85,30 @@ function Show-Karma {
...
+        'OpenFile' {
+            try {
+                $Results = Get-Karma @GetParams
+            }
+            catch {
+                $PSCmdlet.ThrowTerminatingError($_)
+            }
+            $UnformatedPath     = $Results.Meditation
+            $PSKoansFilePath = $UnformatedPath.Substring($UnformatedPath.IndexOf(', ')+2,$UnformatedPath.IndexOf(': ')-($UnformatedPath.IndexOf(', ')+2))
+            $PSKoansLine = $UnformatedPath.Substring($UnformatedPath.IndexOf(': line ')+7,$UnformatedPath.LastIndexOf(("`n"))-($UnformatedPath.IndexOf(': line ')+7))
```
vexx32: You're right, there has to be a better way to figure out the current topic path! Not sure about how to jump to the exact line number, that might require something more complex...

Anyway, you can get the file path much easier by doing something like this:

```powershell
$Results = Get-Karma @GetParams

$Path = Get-PSKoan -Topic $Results.CurrentTopic.Name -Scope User |
    Select-Object -ExpandProperty Path
```

As for the line number... since the meditation message is split over two lines always, you can grab the second line and remove everything after the colon:

```powershell
($Results.Meditation -split '\r?\n')[1] -replace ':.+'
```

vexx32: Make sure you add comments to the latter one, it's a bit of a mouthful of code, in the end. 

Perhaps we could also alter `Get-Karma` to add the line number and full topic path to the `CurrentTopic` properties, making this part trivial? The information `Get-Karma` has access to should allow us to get the line number a bit more easily, I think, and it definitely has the path information available, or we could just get it the same way and store it from the beginning.

That would also make it easier for other folks to write their own Show-Karma style wrappers that have similar capabilities.

PSKoans/Public/Show-Karma.ps1
arefg: I think "-Contemplate -File" is a little bit too complicated and hard to find for such a useful switch, don't you think?  We can even merge these 2 into just Contemplate, after all there is no need to just open the folder when you can open the file and jump to the line. 😉

PSKoans/Public/Show-Karma.ps1
```diff
@@ -80,6 +85,29 @@ function Show-Karma {
...
+        'OpenFile' {
+            try {
+                $Results = Get-Karma @GetParams
+            }
+            catch {
+                $PSCmdlet.ThrowTerminatingError($_)
+            }
+            $FilePath = Get-PSKoan -Topic $Results.CurrentTopic.Name -Scope User | Select-Object -ExpandProperty Path
+            $LineNumber = $Results.CurrentTopic.FailedLineNumber
+            if ($PSVersionTable.Platform -eq "win32nt") {
```
arefg: This works great in Windows, but we need to do the same for Linux, for now, I just open the file with the default editor, any suggestion is appreciated. 😊

PSKoans/Public/Get-Karma.ps1
```diff
@@ -102,17 +102,18 @@
...
+                    It                      = $NextKoanFailed.Name
+                    Expectation             = $NextKoanFailed.ErrorRecord
+                    Meditation              = $NextKoanFailed.StackTrace
+                    KoansPassed             = $KoansPassed
+                    TotalKoans              = $TotalKoans
+                    CurrentTopic            = [PSCustomObject]@{
+                            Name                = $KoanFile.Topic
+                            Completed           = $PesterTests.PassedCount
+                            Total               = $PesterTests.TotalCount
+                            FailedLineNumber    = ($NextKoanFailed.StackTrace -split '\r?\n')[1] -replace ':.+' #Grabs the line number of the failed test from the second line of the StackTrace
```
vexx32:
```suggestion
                            #Grabs the line number of the failed test from the second line of the StackTrace
                            FailedLineNumber    = ($NextKoanFailed.StackTrace -split '\r?\n')[1] -replace ':.+'
```

PSKoans/Public/Show-Karma.ps1
vexx32: Hmm. That's a good point, it may be worth doing that.

PSKoans/Public/Show-Karma.ps1
vexx32: Hmm. Well, we can't use this on Windows PowerShell; `Platform` was added sometime in PS6+

Also, this won't respect the `Editor` setting from PSKoans itself (`Get-PSKoanSetting -Name Editor`); that should probably be all we check tbh. It's possible we should have a more flexible / system-aware default setting, but I think that would be a separate PR entirely and would be taken care of outside this command. All this needs to do is query the selected editor, which the user can set at any time with `Set-PSKoanSetting`.

PSKoans/Public/Show-Karma.ps1
```diff
@@ -80,6 +85,29 @@ function Show-Karma {
...
+                $Editor = (Get-ItemProperty -Path 'HKCU:\Software\Microsoft\Windows\CurrentVersion\Explorer\FileExts\.ps1\UserChoice').ProgId
+                $Editor =  [regex]::matches($DefaultProgram,'(?<=\\).+(?=.exe)').value
+                switch ($Editor) {
+                    {($_ -eq "code") -or ($_ -eq "code-insiders")} {$FullArgument = '-g "{0}":{1}' -f $FilePath, $LineNumber; Start-Process $Editor -ArgumentList $FullArgument}
```
vexx32:
```suggestion
                    { $_ -in 'code', 'code-insiders' } {
                        $FullArgument = '--goto "{0}":{1} --reuse-window' -f $FilePath, $LineNumber
                        Start-Process $Editor -ArgumentList $FullArgument
                    }
```
Bit too much going on here for one line, I think! 😄 Probably doesn't hurt to use full parameter names for VS code so it's a bit easier on the eyes as well...

PSKoans/Public/Show-Karma.ps1
```diff
@@ -80,6 +85,29 @@ function Show-Karma {
...
+                    atom {$FullArgument = '"{0}":{1}' -f $FilePath, $LineNumber; Start-Process $Editor -ArgumentList $FullArgument}
```
vexx32:
```suggestion
                    atom {
                        $FullArgument = '"{0}":{1}' -f $FilePath, $LineNumber
                        Start-Process $Editor -ArgumentList $FullArgument
                    }
```

PSKoans/Public/Show-Karma.ps1
```diff
@@ -80,6 +85,29 @@ function Show-Karma {
...
+                    powershell_ise {$FullArgument = "{0}" -f $FilePath; Start-Process $Editor -ArgumentList $FullArgument}
```
vexx32:
```suggestion
                    {$_ -in 'powershell_ise', 'isep' } {
                        $FullArgument = "{0}" -f $FilePath
                        Start-Process -FilePath $Editor -ArgumentList $FullArgument
                    }
```

Should probably cover the ISE-preview editor here as well. Probably not a lot of folks using it, but it doesn't hurt to be careful.

PSKoans/Public/Show-Karma.ps1
vexx32: I don't see why this wouldn't work on Unix OSes? Starting a process generally works about the same on both OSes, minus some edge cases with some issues around quoting command line arguments... but I don't think we're going to hit those there.

PSKoans/Public/Show-Karma.ps1
arefg: Yep this is much better 😄

PSKoans/Public/Show-Karma.ps1
arefg: I was talking about the registry line. 😊 I don't have much Linux experience but for sure the registry key extraction of the default editor name won't work in Linux. So probably there should be a file or something to extract the default program name for a specific extension in Linux.

vexx32: Yeah, that's not something I'd expect this command to handle, I don't think. We can look into possibly examining that when we set the initial default setting, but I don't think we should be overriding a setting that may come from the user here. 🙂

arefg: There is another problem that I couldn't figure out how to resolve. Atom, VSCode, PowerShell_ISE they all have a registry key or environment variable I think that makes them easy to run (by just typing their name in the terminal) but this is not the case with "Sublime Text 3" and some other editors. Either we need to make them run like the others by adding a key or variable or use a completely different way to be able to cover more IDEs. Any suggestion? 🤔

vexx32: That's not something we can handle, really. The editors that do work are adding the path to the folder their executables are in to the $env:PATH variable; if folx want to do that themselves they can make that work with [Environment]::SetEnvironmentVariable(), but tbh messing with the path variables is something that we probably shouldn't delve into here. 

We can always add something to the documentation to explain how to add folders to the PATH variable if it becomes an issue. 🙂

PSKoans/Public/Show-Karma.ps1
```diff
@@ -80,6 +85,38 @@ function Show-Karma {
...
+                $Results = Get-Karma @GetParams
+            }
+            catch {
+                $PSCmdlet.ThrowTerminatingError($_)
+            }
+            $FilePath = Get-PSKoan -Topic $Results.CurrentTopic.Name -Scope User | Select-Object -ExpandProperty Path
+            $LineNumber = $Results.CurrentTopic.FailedLineNumber
+            if (($PSVersionTable.PSVersion.Major -le 5) -or ($PSVersionTable.Platform -eq "win32nt")) {
+                $Editor = (Get-ItemProperty -Path 'HKCU:\Software\Microsoft\Windows\CurrentVersion\Explorer\FileExts\.ps1\UserChoice').ProgId
+                $Editor =  [regex]::matches($DefaultProgram,'(?<=\\).+(?=.exe)').value
```
vexx32: We shouldn't be getting the editor from registry. It's fine to set this as a default (probably), but we don't handle defaults here. All we need do in this function is grab the _current_ setting from the module settings with `Get-PSKoanSetting -Name Editor` and use that. 🙂 

We don't need to handle platform or any of those things here. If we want to do something like this, that'd be handled when the module is first imported, and we can figure that out in a separate PR.

PSKoans/Public/Show-Karma.ps1
vexx32: I think we're best off refactoring the current implementation into `-Contemplate` itself, and then we can either remove the extra switch or have an `-OpenFolder` instead, which uses the current `-Contemplate` behaviour so users can still do that as well. 🙂

PSKoans/Public/Show-Karma.ps1
arefg: Now I'm confused 😄 
If I'm not mistaken, Get-PSKoanSetting reads the editor from a config file which has a hardcoded value for the editor, do you want to detect the editor and put it in the config file and then read it from there? Should we put the detecting code in the Set-PSKoanSetting or somewhere else? Do you have a better way of detecting the editor other than reading the registry?

vexx32: None of this should be handled in this command 😄 

If we choose to look for some default file association, that should be determined well before Show-Karma is ever called. 🙂 

All this command needs to do is look for the current setting and go from there.

The only place we'd look for this would be a one-time lookup when the module is first imported and the initial settings are created, I think. There's not much point in having an `Editor` setting that users can change if we aren't respecting it here. 😁

arefg: I put it in Show-Karma to be able to detect any change in the default editor, if we don't check the current default program every time we want to open a file we may end up opening the file in a non-default editor(user may change the default program in windows for .ps1 extension after importing the module). Either we need to check it in Show-Karma or call a function to get the latest result.  Am I right? 🤔

vexx32: The user may prefer Show-Karma to use a specific editor for any number of reasons. I don't think it makes sense to determine that dynamically. Even on windows, file associations are not necessarily reliable as they can be modified by other applications.

I would prefer one-time detection and then if the user wants to change their editor we should allow them to set it manually and have PSKoans respect that setting.

arefg: Ok no detection here, registry things removed 😉 
Are we good?

PSKoans/Public/Show-Karma.ps1
arefg: All done

arefg:
Hey Joel, I know this is a stupid question 😄  but since I'm new to all these Github stuff, what should I do now? Should I click on "Close and comment"? :stuck_out_tongue:

PSKoans/Public/Show-Karma.ps1
```diff
@@ -80,6 +84,32 @@ function Show-Karma {
...
+                }
+                atom {
+                    $FullArgument = '"{0}":{1}' -f $FilePath, $LineNumber
+                    Start-Process $Editor -ArgumentList $FullArgument
+                }
+                {$_ -in 'powershell_ise', 'isep' } {
+                    $FullArgument = "{0}" -f $FilePath
+                    Start-Process -FilePath $Editor -ArgumentList $FullArgument
+                }
+                Default {Invoke-Item $FilePath}
```
vexx32: 🤔 We should probably have the default here still call $Editor if there's anything in that, so maybe something like this?

```suggestion
                default {
                    if ($Editor) {
                        $FullArgument = "{0}" -f $FilePath
                        & $Editor $FullArgument
                    }
                    else {
                        Invoke-Item $FilePath
                    }
                }
```

vexx32:
@arefg once you update the tests and everything in the CI is green I'll merge the PR, no worries :)

vexx32:
I'll continue this in #346 for now. Thanks for the great work so far! 😊 💖

PR #319, opened 2019-10-30
Publish External Help with module

vexx32:
﻿# PR Summary

At some point during CI refactoring the external help XML stopped being published with the built module and thus is missing from the nupkg on the gallery.

## Changes

- Update CI and publish steps to make sure the module artefacts are published with the help XML.
- Add a check in the publish step to verify that the help XML is present, and fail the publish step if it is not.

## Checklist

- [x] Summarised changes.


vexx32:
Verified that the module artifact and nupkg artifact include the en-us help XML properly.

PR #368, opened 2020-03-04
Update Readme.md - Installation of Pester Module

checkso:
If windows 10 v1809 is used the installation of Pester is not working as described and needs to be handled like in the suggestion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,16 @@ Install-Module Pester -Force -SkipPublisherCheck -Scope CurrentUser
 Install-Module Pester -Scope CurrentUser
 ```
 
+In Windows 10 v1809 and higher, you first need to cleanup the default Pester module and only then you can proceed with the installation of higher version of Pester module
+
+```$module = "C:\Program Files\WindowsPowerShell\Modules\Pester"
+takeown /F $module /A /R
+icacls $module /reset
+icacls $module /grant "*S-1-5-32-544:F" /inheritance:d /T
+Remove-Item -Path $module -Recurse -Force -Confirm:$false
+Install-Module -Name Pester -Force
+```
+
 ## Getting Started
 
 ### Install from Gallery


vexx32:
Does the install command above with -Force and additional parameters not work?

There shouldn't be a need to completely remove the in-box installation in my experience :slightly_smiling_face:

checkso:
Didn't work for me. 

And this is from the official installation documentation of pester.

vexx32:
No worries, I'll have to get a clean VM spun up so I can test whether the existing docs are accurate or not. 🙂

vexx32:
@checkso I just checked on a fresh Win10 VM that upgrading Pester in the way currently documented in the installation steps here does work. 

I'm not sure what issue you ran into here, but I don't think it's a common one. If you can determine what exactly caused it, I'd be happy to add some cautionary warnings to the docs, but I'm not sure that clearing out the original Pester install is the best course of action.

I might chat to @nohwnd at some point and see if he's seen some things I'm not aware of presently that might need a mention here as well.

Thanks for bringing this up nonetheless! 💖 😊 🌸

checkso:
@vexx32 sure no problem.
Interesting that it worked for you. Your machine was on Windows 10 1809? 
Because from the official documentation it looks like this is the only way to install a newer version of pester: https://pester.dev/docs/introduction/installation/

"In Windows 10 v1809 and higher, you first need to cleanup the default Pester module and only then you can proceed with the installation of higher version of Pester module"

nohwnd:
I think our docs are outdated there is exclusion (for along time) in the recent version of PowershellGet to allow upgrade from the MS signed module to the one signed by us. It is still a good idea to remove the built-in old version.

vexx32:
I'd tend to agree, but I can never be sure if folks still have things using an old pester version specifically, so I'm wary of flat out recommending that it always be removed. 🙂

PR #387, opened 2020-05-10
Small fixes for typos and clarity

devlincashman:
# PR Summary

## Context
I've been completeling the Koans and along the way trying to fix any errors I happen to catch. Most of the changes I've done are for broken tests, but I did change a few things for clarity that caught me up. 

I asked in the #bridge PS channel about a few of these like the sort method for the kata, so if my changes were incorrect there sorry about that.

Also wanted to say thanks so much for all the hard work you do on this project Joel! And of course everyone else who contributes.

See you in chat!

## Changes

- 

- Updated TestDrive: to TEMP: in PSProviders as TestDrive: is not a valid drive from what I could tell?

- Changed several arrays, Pester tests that were failing and seemed likely to be an error.

- Sorting Kata has capitals in the sorted characters at different points, so a simple sort gets an error. Assuming you are not asking for a merge sort which seems advanced. Simplified test strings to pass a basic sort.

## Checklist

- [ ] Pull Request has a meaningful title.
- [ ] Summarised changes.
- [ ] Pull Request is ready to merge & is not WIP.

### DIFF
--- a/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
@@ -134,7 +134,7 @@ Describe 'Environment Provider' {
 
 Describe 'FileSystem Provider' {
     BeforeAll {
-        $Path = 'TestDrive:' | Join-Path -ChildPath 'File001.tmp'
+        $Path = 'TEMP:' | Join-Path -ChildPath 'File001.tmp'
 
         $FileContent = @'
 PSKOANS!
--- a/PSKoans/Koans/Foundations/AboutHashtables.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutHashtables.Koans.ps1
@@ -180,7 +180,7 @@ Describe 'Hashtables' {
             $Hashtable.ContainsValue('Fruit') | Should -BeTrue
 
             $Hashtable['Oranges'] | Should -Be 'Fruit'
-            $Hashtable['Carrots'] | Should -Not -Be $Hashtable['Oranges']
+            $Hashtable['Carrots'] | Should -Be $Hashtable['Oranges']
         }
 
         It 'will not implicitly convert keys and lookup values' {
--- a/PSKoans/Koans/Foundations/AboutLoopsAndPipelines.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutLoopsAndPipelines.Koans.ps1
@@ -125,7 +125,7 @@ Describe 'Pipelines and Loops' {
                 #>
                 $i
             }
-            $Values | Should -Be @(0, 1, 2, 3, 4)
+            $Values | Should -Be __
         }
 
         It 'can loop while a condition is $true' {
--- a/PSKoans/Koans/Katas/SortingCharacters.Koans.ps1
+++ b/PSKoans/Koans/Katas/SortingCharacters.Koans.ps1
@@ -67,11 +67,11 @@ Describe 'Kata - Sorting Characters' {
         }
         @{
             String = 'What do you call the world?'
-            Result = 'aacddehhlllooorttuwWy'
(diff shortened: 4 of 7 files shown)


PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
```diff
@@ -134,7 +134,7 @@ Describe 'Environment Provider' {
 
 Describe 'FileSystem Provider' {
     BeforeAll {
-        $Path = 'TestDrive:' | Join-Path -ChildPath 'File001.tmp'
+        $Path = 'TEMP:' | Join-Path -ChildPath 'File001.tmp'
```
vexx32, line 137: `TEMP:` as a pre-existing PSDrive doesn't exist until PS6 or 7 I think?

`TestDrive:` is provided by Pester and cleans up after itself without us having to do any manual work to do so. Was this one not working for you? 🙂

PSKoans/Koans/Foundations/AboutHashtables.Koans.ps1
```diff
@@ -180,7 +180,7 @@ Describe 'Hashtables' {
             $Hashtable.ContainsValue('Fruit') | Should -BeTrue
 
             $Hashtable['Oranges'] | Should -Be 'Fruit'
-            $Hashtable['Carrots'] | Should -Not -Be $Hashtable['Oranges']
+            $Hashtable['Carrots'] | Should -Be $Hashtable['Oranges']
```
vexx32, line 183: This might need more clarification, for sure, but I don't think this is the problem here. 

The conditions as set without changing this line are that the hashtable must:

1. contain at least one key name that is `Carrots`
1. have at least one of the values be `'Fruit'`
1. the key `Oranges` should map to the value `'Fruit'`
1. the value for `Carrots` should not be the same as the value for `Oranges`

The intended solution is for carrots to be marked as `'Vegetables'`  when adding keys to the hashtable. 🙂 

Is there a way we can make that a bit more clear for this koan?

PSKoans/Koans/Foundations/AboutLoopsAndPipelines.Koans.ps1
```diff
@@ -125,7 +125,7 @@ Describe 'Pipelines and Loops' {
                 #>
                 $i
             }
-            $Values | Should -Be @(0, 1, 2, 3, 4)
+            $Values | Should -Be __
```
vexx32, line 128: Hmm. I tend to prefer that blanks take a similar form to what's expected; we're expecting an array here, so a single blank is a bit misleading imo.

Perhaps something like this would be better?
```ps
$ExpectedValues = @(
    __
    __
    __
    __
    __
)

$ExpectedValues | Should -Be $Values
```

PSKoans/Koans/Katas/SortingCharacters.Koans.ps1
```diff
@@ -67,23 +67,23 @@ Describe 'Kata - Sorting Characters' {
...
         }
         @{
             String = 'The hands of the world are open.'
-            Result = 'aaddeeeefhhhlnnoooprrstTw'
+            Result = 'aaddeeeefhhhlnnoooprrsTtw'
         }
         @{
             String = 'You are those huge waves sweeping everything before them, swallowing all in their path.'
-            Result = 'aaaaabeeeeeeeeeeeefgggghhhhhhiiiiillllmnnnnoooopprrrrsssstttttuuvvwwwwyY'
+            Result = 'aaaaabeeeeeeeeeeeefgggghhhhhhiiiiillllmnnnnoooopprrrrsssstttttuuvvwwwwYy'
```
vexx32, line 86: Can you share the code you used to get these results?

I remember what I did to get the result I laid out in there, but there's a good chance it didn't play nice with the upper and lower cases. This looks right, just want to verify :)

vexx32:
@devlincashman are you still looking to continue / finish this one up? 🙂

PR #399, opened 2020-06-22
Fix Issue with Modules Koans

vexx32:
# PR Summary

There seems to be a few issues with the module koans not running correctly. This PR should resolve those problems.

## Context

Original issues reported by @shaneis on Discord.

## Changes

- Get-PSKoan - Don't write errors when the `Modules` folder is missing; Get-Karma will take the correct action and cause them to be updated anyway, so the error is not useful.
- Invoke-Koan - Check koan script for any requirements before running it and automatically import required modules in the new runspace before invoking Pester.

## Checklist

- [x] Summarised changes.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,36 +5,18 @@
     "version": "0.2.0",
     "configurations": [
         {
-            "name": "PowerShell Launch Current File",
+            "name": "PowerShell Launch Script",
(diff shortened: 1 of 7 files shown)


vexx32:
Tests are required for this, but will have to wait till #398 because good _lord_ I do not wanna go there with that merge conflict it'll cause.

PR #415, opened 2020-10-05
Fix 'can be used to select a value for a variable'

### LINKED ISSUE #414: Koan uses file length that can change breaking Koan

Foundations/AboutConditionals.Koans.ps1 - 'can be used to select a value for a variable'

```powershell
function Get-Thing {
    return (Get-ChildItem -Path $HOME -File | Select-Object -First 1).Length
}
```

## Context "Your Attempts"

Koan will pass but running it a different time after the length has changed breaks the Koan.

The Koan should not use information that is dynamic on the system as the value can change causing the Koan to pass or fail randomly.

ThomasNieto:
# PR Summary

Koan Foundations/AboutConditionals.Koans.ps1 - 'can be used to select a value for a variable'

Fixes #414 

## Context

This PR updates the Koan to create a test file in Pester TestDrive so the Koan no longer uses dynamic system information.

## Changes

* Refactor to use Pester `TestDrive`

## Checklist

- [x] Summarised changes.

### DIFF
--- a/PSKoans/Koans/Foundations/AboutConditionals.Koans.ps1
+++ b/PSKoans/Koans/Foundations/AboutConditionals.Koans.ps1
@@ -29,7 +29,7 @@ Describe 'If/Else' {
             }
 
             $ExpectedResult = Assert-IsEven -Number 2
-            '____' | Should -Be $ExpectedResult
+            'EVEN' | Should -Be $ExpectedResult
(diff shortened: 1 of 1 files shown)


vexx32 (approved):
Looks good to me. Thanks for fixing this! 💖 😊

Issue #260, opened 2019-09-25
Need Advice Snippets


vexx32:
# Describe "Functionality"

https://github.com/vexx32/PSKoans/tree/master/PSKoans/Data/Advice

Currently there are 3 available snippets for the cmdlets to pull from. We need a lot more. Consider this a meta issue for tossing ideas into and/or asking for perspective on ideas; PRs for individual advice snippets or entire collections are more than welcome!

## Context "Additional Information"

Advice snippets should be reasonably short, informative, and self-contained. They can be about PowerShell, useful modules for various purposes, general code practices, common pitfalls and how to recognise and avoid them, etc., etc.

Any amount of helpful advice for newbies and intermediate-level folks is more than welcome. 😊

PR #420, opened 2020-10-26
Profile.Advice.json: Expand on $profile paths

Maamue:
# PR Summary

Expand on Profile Advice as asked in #260 

## Context

- expand on $profile possibilities (different PSHosts)
- expand on changes IDEs make on $profile

## Changes

only textual changes with different possibilities where $profile can point to

## Checklist

- [x] Summarised changes.

### DIFF
--- a/PSKoans/Data/Advice/Customization/Profile.Advice.json
+++ b/PSKoans/Data/Advice/Customization/Profile.Advice.json
@@ -4,6 +4,16 @@
     "When starting powershell, there is a script it will automatically execute: the PowerShell profile script.",
     "By default, it does not exist. It is stored in the location the variable $profile points to. By creating and adding code to it you can ensure that all your favorite aliases, variables, functions or whatnot will automatically be available each time!",
     "To create or open the profile file, simply run:",
-    "  notepad $profile"
+    "  notepad $profile",
+    "\n",
+    "The $profile variable stores the path to the 'Current User, Current Host' profile, that points to a Microsoft.PowerShell_profile.ps1 file. Other profiles can be retreived from note properties of the $profile variable:",
+    "AllUsersAllHosts",
+    "AllUsersCurrentHost",
+    "CurrentUserAllHosts",
+    "CurrentUserCurrentHost",
+    "\n",
+    "Different PSHosts (applications hosting a powershell process) like IDEs have custom profiles paths that are referenced by $profile. While the default $profile will be available in every environment, it is useful to know where IDEs save their profiles. Customizations only required for e.g. development can be added.",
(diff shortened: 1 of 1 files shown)


PSKoans/Data/Advice/Customization/Profile.Advice.json
```diff
@@ -4,6 +4,16 @@
...
-    "  notepad $profile"
+    "  notepad $profile",
+    "\n",
+    "The $profile variable stores the path to the 'Current User, Current Host' profile, that points to a Microsoft.PowerShell_profile.ps1 file. Other profiles can be retreived from note properties of the $profile variable:",
+    "AllUsersAllHosts",
+    "AllUsersCurrentHost",
+    "CurrentUserAllHosts",
+    "CurrentUserCurrentHost",
+    "\n",
+    "Different PSHosts (applications hosting a powershell process) like IDEs have custom profiles paths that are referenced by $profile. While the default $profile will be available in every environment, it is useful to know where IDEs save their profiles. Customizations only required for e.g. development can be added.",
```
vexx32, line 15:
```suggestion
    "Different PSHosts (applications hosting a powershell process) like IDEs have custom profile paths that are referenced by $profile. While the default $profile will be available in every environment, it is useful to know where IDEs save their profiles. Customizations only required for e.g. development can be added.",
```

PSKoans/Data/Advice/Customization/Profile.Advice.json
```diff
@@ -4,6 +4,16 @@
...
+    "Powershell ISE uses $profile, but with a custom path pointing to a Microsoft.PowerShellISE_profile.ps1 file.",
+    "Visual Studio Code uses $profile, but with a custom path pointing to a Microsoft.VSCode_profile.ps1 file."
```
vexx32, line 17: I think we should clarify these custom paths are incorporated into the "current host" variants of the profile paths.

vexx32 (changes requested):
Thanks for taking care of this! 💖 

There's one or two small things that I think we should alter a little but other than that, this looks fantastic. Thanks! 😊 🌸

PSKoans/Data/Advice/Customization/Profile.Advice.json
Maamue: Good catch! Added another sentence to explain the same behaviour.

PSKoans/Data/Advice/Customization/Profile.Advice.json
Maamue: should be fixed now, thanks :)

vexx32 (approved):

PR #427, opened 2021-02-20
AboutCmdletVerbs: Use Join-Path for hardcoded /

Maamue:
# PR Summary

Change hardcoded forward slash / in filepath variable into os-dependent path separator with Join-Path

## Context

$FilePath was declared "$env:TMP/YOUR_PATH.txt". This leads to an issue on Windows platforms on the last pester test requesting an error message. While
Powershell itself treats both / and \ without any contempt, the error message expects backslashes for the path. Without Join-Path and its os dependant path separator, the test for the error message can't be fulfilled. The path it checks against is given as (shorthand)

AppData\Local\Temp/YOUR_PATH.txt

due to the koan hardcoding the forward slash into the variable. With Join-Path building the path, it gets resolved correctly as

AppData\Local\Temp\YOUR_PATH.txt

## Changes
"$env:TMP/YOUR_PATH.txt" --> Join-Path -Path "$env:TMP" -ChildPath "YOUR_PATH.txt"

## Checklist

- [x] Summarised changes.

### DIFF
--- a/PSKoans/Koans/Introduction/AboutCmdletVerbs.Koans.ps1
+++ b/PSKoans/Koans/Introduction/AboutCmdletVerbs.Koans.ps1
@@ -22,7 +22,7 @@ Describe "Basic Verbs" {
 
     BeforeAll {
         # We'll be using this path later on.
-        $FilePath = "$env:TMP/YOUR_PATH.txt"
+        $FilePath = Join-Path -Path "$env:TMP" -ChildPath "YOUR_PATH.txt"
 
         if (Test-Path $FilePath) {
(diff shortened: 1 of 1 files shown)


Maamue:
> 
> should be using `$TestDrive` instead of `$env:TMP`?

Changed to `$TestDrive`. I can change it back if necessary, but I agree that `$TestDrive` is to be preferred. We do not want to leave files behind on the filesystem.

vexx32 (approved):

vexx32:
Thanks for taking care of this! 💖

PR #436, opened 2021-04-23
Fix koan AboutPsProviders

### LINKED ISSUE #433: Temp: drive is not available in PowerShell 5.1

```powershell
Show-Karma -Topic AboutpsProviders
```
with PowerShell 5.1 it gets following error:

```
Invoke-Koan : Cannot find drive. A drive with the name 'TEMP' does not exist.
At C:\Program Files\WindowsPowerShell\Modules\PSKoans\0.67.1\Public\Get-Karma.ps1:111 char:32
+                 $PesterTests = Invoke-Koan @{
+                                ~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (TEMP:String) [Invoke-Koan], DriveNotFoundException
    + FullyQualifiedErrorId : DriveNotFound,Invoke-Koan
```

https://github.com/vexx32/PSKoans/blob/c1ad5c98d63d0dbe02a50d4dd56d6a7fd5923810/PSKoans/Koans/Cmdlets%201/AboutPSProviders.Koans.ps1#L137 uses PSDrive TEMP: that is not defined in PowerShell 5.1.

Instead, it could use New-TemporaryFile cmdlet.  That should work both in PowerShell 5.1 and 7.x

### LINKED ISSUE #434: Undesired error with unresolved koan AboutPSProviders

```powershell
Show-Karma -Topic AboutPSProviders
```
emits following error:
```
Invoke-Koan: C:\Program Files\WindowsPowerShell\Modules\PSKoans\0.67.1\Public\Get-Karma.ps1:111
Line |
 111 |                  $PesterTests = Invoke-Koan @{
     |                                 ~~~~~~~~~~~~~~
     | This command cannot find a matching alias because an alias with the definition '____' does not exist.
```
The error appears with both PowerShell 7 and 5.1

The problem is with line https://github.com/vexx32/PSKoans/blob/c1ad5c98d63d0dbe02a50d4dd56d6a7fd5923810/PSKoans/Koans/Cmdlets%201/AboutPSProviders.Koans.ps1#L78

as soon as previous line gets correct cmdlet name, the error ceases to exist.

It would be good idea to add -ErrorAction SilentlyContinue to Get-Alias on that line.

### LINKED ISSUE #435: Bug in koan AboutPSProviders

As soon as It block 'allows access to variables in the current scope' becomes resolved, the next block throws an error:
```
Invoke-Koan: C:\Program Files\WindowsPowerShell\Modules\PSKoans\0.67.1\Public\Get-Karma.ps1:111
Line |
 111 |                  $PesterTests = Invoke-Koan @{
     |                                 ~~~~~~~~~~~~~~
     | Cannot remove variable Test because the variable has been optimized and is not removable. Try using
     | the Remove-Variable cmdlet (without any aliases), or dot-sourcing the command that you are using to
     | remove the variable.
```
The error occurs in both PowerShell 5.1 and 7.1

The error occurs on line: https://github.com/vexx32/PSKoans/blob/c1ad5c98d63d0dbe02a50d4dd56d6a7fd5923810/PSKoans/Koans/Cmdlets%201/AboutPSProviders.Koans.ps1#L261

## Context "Your Attempts"

Renaming variable to something else doesn't help.  But it helps, when the used variable name is conditionally deleted (if exists) in the beginning of the It block.  Something like:
```powershell
if (Test-Path -Path 'Variable:Test') {
    Remove-Variable -Name 'Test'
}
```

It seems to be a scoping issue, but can't understand, from where the variable gets inserted.  Especially when unique name is assigned to variable.

peetrike:
# PR Summary

Fixes 3 problems in koan AboutPsProviders

* Problem with undersired error message when alias placeholder is no yet replaced (fixes #434)
* Problem with Temp: drive not available in PowerShell 5.1 (fixes #433)
* Problem with variable becoming un-deletable (fixes #435)

## Checklist

- [x] Summarised changes.

### DIFF
--- a/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
@@ -134,6 +134,10 @@ Describe 'Environment Provider' {
 
 Describe 'FileSystem Provider' {
     BeforeAll {
+        if (-not (Test-Path -Path Temp: -PathType Container)){
+            # In PowerShell 5.x the Temp: drive is not defined by default
+            New-PSDrive -Name Temp -PSProvider FileSystem -Root $env:TEMP
+        }
         $Path = 'TEMP:' | Join-Path -ChildPath 'File001.tmp'
 
         $FileContent = @'
--- a/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
@@ -258,7 +258,7 @@ Describe 'Variable Provider' {
         }
 
         It 'allows you to remove variables' {
-            Set-Variable -name Test -Value 123
+            Set-Variable -Name Test -Value 123
 
             __ | Should -Be $Test
 
--- a/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
@@ -75,7 +75,7 @@ Describe 'Alias Provider' {
 
         It 'can seek out aliases for a command' {
             $CmdletName = '____'
-            $AliasData = Get-Alias -Definition $CmdletName
+            $AliasData = Get-Alias -Definition $CmdletName -ErrorAction SilentlyContinue
 
             $AliasData.Name | Should -Be 'gcm'
         }
--- a/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
@@ -268,6 +268,9 @@ Describe 'Variable Provider' {
         }
 
         It 'exposes data from default variables' {
+            if (Test-Path -Path 'Variable:Test') {
+                Remove-Variable -Name Test
+            }
             $Variables = Get-ChildItem -Path 'Variable:'
 
             '____' | Should -Be $Variables.Where{$_.Name -eq 'ConfirmPreference'}.Value
--- a/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
@@ -260,7 +260,7 @@ Describe 'Variable Provider' {
         It 'allows you to remove variables' {
             Set-Variable -name Test -Value 123
 
-            123 | Should -Be $Test
+            __ | Should -Be $Test
 
             Remove-Item -Path 'Variable:\Test'
             $____ | Should -Be $Test
--- a/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
@@ -258,6 +258,9 @@ Describe 'Variable Provider' {
         }
 
         It 'allows you to remove variables' {
+            if (Test-Path -Path 'Variable:Test') {
+                Remove-Variable -Name Test
+            }
             $Test = 123
 
             __ | Should -Be $Test
@@ -268,9 +271,6 @@ Describe 'Variable Provider' {
         }
 
         It 'exposes data from default variables' {
-            if (Test-Path -Path 'Variable:Test') {
-                Remove-Variable -Name Test
-            }
             $Variables = Get-ChildItem -Path 'Variable:'
 
             '____' | Should -Be $Variables.Where{$_.Name -eq 'ConfirmPreference'}.Value
--- a/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
+++ b/PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
@@ -258,12 +258,9 @@ Describe 'Variable Provider' {
         }
 
         It 'allows you to remove variables' {
-            if (Test-Path -Path 'Variable:Test') {
-                Remove-Variable -Name Test
-            }
-            $Test = 123
+            Set-Variable -name Test -Value 123
 
-            __ | Should -Be $Test
+            123 | Should -Be $Test
 
             Remove-Item -Path 'Variable:\Test'
             $____ | Should -Be $Test


PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
```diff
@@ -264,6 +268,9 @@ Describe 'Variable Provider' {
         }
 
         It 'exposes data from default variables' {
+            if (Test-Path -Path 'Variable:Test') {
+                Remove-Variable -Name Test
+            }
```
vexx32, line 273: This feels like an odd place for this to go. Does it still work if we remove the variable at the end of `allows access to variables in the current scope` instead?

vexx32 (changes requested):
Looks pretty solid to me, one of these we probably need to adjust a bit and make sure it's in the right place, but other than that these are good. Thanks for doing these! 💖 😊

PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
peetrike: somehow after adding that code, it doesn't complain after adding number to first placeholder.  It's more like a workaround until we understand the real problem and can fix that

peetrike: maybe adding a comment about that code being workaround would help keeping the attention :)

peetrike: I also tried to add the same code to the end of previous It block, but it didn't help at all.  Maybe it isn't about scoping and is realated to something else.

OTOH, the same problem didn't occur in command line.

vexx32: Hmm, yeah. I'll take a closer look today and see if I can find out what's up with this one. 🙂

peetrike: Sorry, i misplaced the block, will fix that

vexx32:
@peetrike I had a look at it, and I think the "neatest" way around the error here is to use `Set-Variable` to create the variable instead of using the normal syntax. That seems to have a similar effect, and doesn't require adding the conditional `Remove-Variable` that you have in your current solution. I think that will be less confusing to users in the long run.

Something in the way PSKoans is invoking that code is causing PowerShell to optimize that code path in certain cases. @seeminglyscience could probably explain what on earth is going on with it, but it's a bit beyond me as to exactly why it's occurring.

peetrike:
will change the last solution

PSKoans/Koans/Cmdlets 1/AboutPSProviders.Koans.ps1
```diff
@@ -254,7 +258,7 @@ Describe 'Variable Provider' {
         }
 
         It 'allows you to remove variables' {
-            $Test = 123
+            Set-Variable -name Test -Value 123
```
vexx32, line 261:
```suggestion
            Set-Variable -Name Test -Value 123
```

vexx32 (approved):
Thanks for fixing this up! 💖 🌸

SeeminglyScience:
> Something in the way PSKoans is invoking that code is causing PowerShell to optimize that code path in certain cases. @SeeminglyScience could probably explain what on earth is going on with it, but it's a bit beyond me as to exactly why it's occurring.

So there are basically two ways a variable is stored internally:

1. in a `Dictionary<string, PSVariable>`.  This is the "unoptimized" route, taken when dot sourcing or when the debugger is active.  Maybe also when code is interpreted rather than JIT compiled
2. In a generated `MutableTuple<..>` that is basically a `ITuple` implementation with a property representing every *local* for the scope

Those errors occur most often when trying to remove a variable from a previous scope that is represented by a mutable tuple entry.

peetrike:
does it mean that in scriptblock, directory assigning to variable makes it "optimized"?  And later when you try to remove it through Variable: drive, it will be prevented?

vexx32:
I've seen it before removing variables from other scopes, but there even if it used a completely different variable name than anything else it still hit it. Really weird.

peetrike:
seems so:
```powershell
& { $MyTest = 11; $MyTest ; Remove-Item variable:MyTest }
```
```
Remove-Item: Cannot remove variable MyTest because the variable has been optimized and is not removable. Try using the Remove-Variable cmdlet (without any aliases), or dot-sourcing the command that you are using to remove the variable.
```

peetrike:
and same in PowerShell 5.1

peetrike:
maybe that could be included as another It block in that Koan?

vexx32:
Sounds like we'll need to, yeah 😁

peetrike:
or maybe it fits better into AboutVariables koan?

vexx32:
I think it's better off in this one, because `Remove-Variable` will still work, it's only the provider access via `variable:` with `Remove-Item` that fails.

SeeminglyScience:
> seems so:
> 
> ```powershell
> & { $MyTest = 11; $MyTest ; Remove-Item variable:MyTest }
> ```

It's less to do with the fact that it's in a scriptblock and more about the invocation operator.  The compiler won't optimize when dot sourcing, which is the default in an interactive session.  So forcing it out of dot sourcing using `&` rather than `.` is what makes the difference there.

vexx32:
Hmm. Do you think there's a way to modify https://github.com/vexx32/PSKoans/blob/main/PSKoans/Private/Invoke-Koan.ps1 to avoid this issue in future?

vexx32:
That would probably be even less clear for users 😁 

Might be good as an example for a different topic though lol. Ah well, at least we can help clarify how that works for folks to some degree 🤔

SeeminglyScience:
Yeah I should have clarified that example is more meant as inspiration than something that should be actually done.  It's pretty incomprehensible even with a lengthy explanation

peetrike:
and that works without problems:
```powershell
& {New-Item -Path variable: -Name Test -Value 11; $Test; Remove-Item variable:Test }
```

peetrike:
not that someone shoud do that :)

SeeminglyScience:
And if you're wondering why that works, it's because the compiler doesn't *know* it's a local.  `New-Item` could be proxied, drive could be replaced (maybe?) etc etc. All it knows is literal variable assignments.

peetrike:
so does it mean that only using assignment in separate scope triggers that "oütimization" of variable?

peetrike:
> Yeah I should have clarified that example is more meant as inspiration than something that should be actually done. It's pretty incomprehensible even with a lengthy explanation

it would be more practical to give people "good" examples, not "bad" ones :)

vexx32:
Honestly my biggest point of confusion here is why does Remove-Variable work when remove-item doesn't? Why can't Remove-Item do what the other cmdlet is doing for that provider?

peetrike:
> Honestly my biggest point of confusion here is why does Remove-Variable work when remove-item doesn't? Why can't Remove-Item do what the other cmdlet is doing for that provider?

probably because `Remove-Variable` is meant to deal with variables, but going through provider makes it different

SeeminglyScience:
> so does it mean that only using assignment in separate scope triggers that "oütimization" of variable?

It needs to not be dot sourced if that's what you mean.

> Honestly my biggest point of confusion here is why does Remove-Variable work when remove-item doesn't?

Either oversight or harder to do with the extra abstraction.  Not sure which it is.

> Why can't Remove-Item do what the other cmdlet is doing for that provider?

Probably can

peetrike:
hm:
```powershell
& {new-variable -name test -Scope local -Value 11; $test2=12; $test2; Remove-Item variable:test2 }
```
```
12
```

SeeminglyScience:
So it looks like the compiler sees a `*variable` command and just nopes out of optimization.  That explains why one works and the other does, and I think you're right @peestrike, can't fix ri the same way.

Especially since this throws:

```powershell
& {& (gcm new-variable) test; $test2=12; $test2; Remove-Item variable:test2 }
```

vexx32:
So it's just completely cursed. Good to know, thanks!

peetrike:
They have been thought about that use case:
```powershell
$ErrorActionPreference = 'SilentlyContinue'
& {$test2=12; Remove-Item variable:test2 }
$error[0].FullyQualifiedErrorId
```
```
VariableNotRemovableRare,Microsoft.PowerShell.Commands.RemoveItemCommand
```

PR #465, opened 2022-01-14
Note that NuGet is only needed on Windows

offby1:
# PR Summary

Note that NuGet is needed only on windows.

## Context

I use MacOS, and tried out the PSKoans; the first thing I tried was installing NuGet, as the README told me to.  That failed like this:
```
PS /Users/erichanchrow> Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force
Install-PackageProvider: No match was found for the specified search criteria for the provider 'NuGet'. The package provider requires 'PackageManagement' and 'Provider' tags. Please check if the specified package has the tags.
```
Someone on discord told me that NuGet wasn't needed for MacOS, and sure enough, I proceeded to try `Install-Module Pester` and `Install-Module PSKoans`; those both worked, and I'm now happily making my way towards enlightenment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,11 @@ View the PSKoans [Command Reference Documentation][reference-docs].
 
 ## Prerequisites
 
-- Windows PowerShell version 5.1 / PowerShell 6+
-- NuGet
+- PowerShell version 5.1 / PowerShell 6+
+- NuGet (Windows only)
 - Pester v4.x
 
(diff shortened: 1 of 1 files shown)


vexx32 (approved):
Ah, good catch, these instructions probably haven't been updated in a while. Thanks for the PR! 💖

PR #471, opened 2022-02-27
Fixes #470

### LINKED ISSUE #470: Regex: $matches[0] is empty within "It does not need to match the entire string" and "It is not just an array of text"

Constructs and Patterns > AboutRegularExpressions

```powershell
    Context '*' {

        It 'specifies 0 or more of something' {
            <#
                The * character specifies "0 or more" of the symbol or group that comes 
                immediately before it. In this case, the pattern can be read aloud as "zero or more
                of the letter p, then the letters e, a, r, and s."
            #>
    
            $____ | Should -Be ('pears' -match 'p*ears')
            $____ | Should -Be ('shears' -match 'p*ears')
        }

        It 'does not need to match the entire string' {
            <#
                What happened in that last example? What part of 'shears' matched the pattern 
                'p*ears'? When you use the -match operator, you can see some interesting information
                in the automatic variable $matches. $matches is a collection of the match objects
                that are created when you use the -match operator. This isn't as flexible nor as
                robust as using [regex]::Matches(), but it's handy in a pinch.
                
                In this case, you'll see that the "sh" in "shears" weren't part of the match. The
                pattern "zero or more p's, followed by e, a, r, s," is found in the string, even
                though there are parts of the string that aren't relevant to that matching effort.
    
                $matches[0] is always your most recent match.
            #>
    
            '____' | Should -Be $matches[0]
        }

        It 'is not just an array of text' {
            # What else do you think $matches might contain?
            '____' | Should -Be $matches.GetType().FullName
        }
    }
```

```powershell
    Describing Quantifiers
      *
        [▸] It specifies 0 or more of something
        [×] It does not need to match the entire string
        [×] It is not just an array of text
```

```
❯ $matches[0]
Cannot index into a null array.
At line:1 char:1
+ $matches[0]
+ ~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : NullArray
```

## Context "Your Attempts"

```powershell
    Context '*' {

        BeforeEach {
            $firstTest = ('pears' -match 'p*ears')
            $secondTest = ('shears' -match 'p*ears')
        }

        It 'specifies 0 or more of something' {
            <#
                The * character specifies "0 or more" of the symbol or group that comes
                immediately before it. In this case, the pattern can be read aloud as "zero or more
                of the letter p, then the letters e, a, r, and s."
            #>

            $____ | Should -Be $firstTest
            $____ | Should -Be $secondTest
        }

        It 'does not need to match the entire string' {
            <#
                What happened in that last example? What part of 'shears' matched the pattern
                'p*ears'? When you use the -match operator, you can see some interesting information
                in the automatic variable $matches. $matches is a collection of the match objects
                that are created when you use the -match operator. This isn't as flexible nor as
                robust as using [regex]::Matches(), but it's handy in a pinch.

                In this case, you'll see that the "sh" in "shears" weren't part of the match. The
                pattern "zero or more p's, followed by e, a, r, s," is found in the string, even
                though there are parts of the string that aren't relevant to that matching effort.

                $matches[0] is always your most recent match.
            #>

            '____' | Should -Be $matches[0]
        }

        It 'is not just an array of text' {
            # What else do you think $matches might contain?
            '____' | Should -Be $matches.GetType().FullName
        }
    }
```

tried adding the Tests into the BeforeEach (breaks the flow a bit, I know). Alternative is to test again within the following two

DEberhardt:
It 'does not need to match the entire string' now has $matches[0] populated without having to test again.

Unsure whether a retest within the following IT statements would be better for flow, but this works.

# PR Summary

## Checklist

- [ ] Pull Request has a meaningful title.
- [ ] Summarised changes.
- [ ] Pull Request is ready to merge & is not WIP.

### DIFF
--- a/PSKoans/Koans/Constructs and Patterns/AboutRegularExpressions.Koans.ps1
+++ b/PSKoans/Koans/Constructs and Patterns/AboutRegularExpressions.Koans.ps1
@@ -156,15 +156,20 @@ Describe 'Quantifiers' {
     
     Context '*' {
 
+        BeforeEach {
+            $firstTest = ('pears' -match 'p*ears')
+            $secondTest = ('shears' -match 'p*ears')
+        }
+
         It 'specifies 0 or more of something' {
             <#
-                The * character specifies "0 or more" of the symbol or group that comes 
+                The * character specifies "0 or more" of the symbol or group that comes
                 immediately before it. In this case, the pattern can be read aloud as "zero or more
                 of the letter p, then the letters e, a, r, and s."
             #>
-    
-            $____ | Should -Be ('pears' -match 'p*ears')
-            $____ | Should -Be ('shears' -match 'p*ears')
+
+            $____ | Should -Be $firstTest
+            $____ | Should -Be $secondTest
         }
 
         It 'does not need to match the entire string' {


vexx32 (approved):
This change looks good to me. Thanks for fixing this up!

PR #493, opened 2026-09-17
build: add standard psake bootstrap (Init/Clean/Build/Test/Analyze/Publish)

### LINKED ISSUE #485: Add standard psake build entry points
## Goal

Add the standard psake bootstrap described by Phase 3b.

## Acceptance criteria

- Root `psakeFile.ps1` exposes `Init`, `Clean`, `Build`, `Test`, `Analyze`, and `Publish`.
- `Invoke-psake ?` lists the tasks.
- `Invoke-psake Test` runs the Pester suite successfully.

HeyItsGilbert:
## Goal

Add the standard psake bootstrap described by Phase 3b of the revival playbook, mirroring PSDepend's `requirements.psd1` + `build.ps1` + `psakeFile.ps1` convention.

Closes #485

## What's here

- `requirements.psd1` / `build.ps1` / `psakeFile.ps1` at repo root, matching PSDepend's psake+PowerShellBuild+PSDepend layout.
- `psakeFile.ps1` uses `-FromModule PowerShellBuild`, adding `Init`, `Clean`, `StageFiles`, `Build`, `Analyze`, `Pester`, `Test`, `BuildHelp`, `Publish`, and `?`. `Invoke-psake ?` lists all tasks.
- `GenerateFormatData` task runs ahead of `StageFiles` since the manifest's `FormatsToProcess` entry (`PSKoans.format.ps1xml`) is EZOut-generated and gitignored — the module can't even import without it.
- `.gitignore` extended for `Output/**`, `Tests/out/**`, `testResults.xml`.

## Test suite bootstrap

Replaced `#Requires -Modules PSKoans` across all 26 `Tests/**/*.Tests.ps1` files with a `BeforeDiscovery` block that builds (if needed) and imports the staged module from `Output/`, matching PSDepend's own staged-import test convention. This surfaced and fixed real, pre-existing bugs:

- **Unscoped mocks** across ~13 test files: without `-ModuleName 'PSKoans'` a `Mock` never intercepts calls made from *inside* the module, so real cmdlets ran instead — including `Start-Process` launching a real editor during a local test run.
- **Data-corruption bug**: `Reset-PSKoan.Tests.ps1` and `Update-PSKoanFile.Tests.ps1` called `(Get-PSKoan -Scope Module).Path` directly from test scope. Module-scoped mocks don't intercept direct test-scope calls, so the real cmdlet ran and `Set-Content` overwrote every staged koan file with fixture content. Fixed by computing paths into variables instead of depending on mock interception outside module scope.
- `requirements.psd1`'s Pester pin relaxed (no exact version): PowerShellBuild's `Test-PSBuildPester` unconditionally does `Import-Module Pester -MinimumVersion 5.0.0`, which throws if a different Pester major is already loaded in-process (reproduced deterministically outside psake).
- Added a `$env:PSModulePath` entry for the staged module root so `using module PSKoans` resolves during koan file parsing/execution — without it, `KoanValidation.Tests.ps1`'s static parse checks failed for every real koan topic.
- Fixed a handful of Pester 6 strictness regressions surfaced by the above (unmatched `-ParameterFilter` mocks, empty `-TestCases` arrays crashing discovery).

## Documentation

Restored comment-based help for all 13 exported functions, ported straight from the existing `docs/*.md` PlatyPS output (parameter sets verified to still match current source — no drift), matching the style already used in `PSKoans/Private/*.ps1`. Fixes the `ModuleHelp.Tests.ps1` documentation-content checks.

## Verification

`./build.ps1 Test` runs clean end to end — no hangs, no crashes: **739 passed / 10 failed**.

All 10 remaining failures trace to one pre-existing, out-of-scope bug: `Invoke-Koan.ps1` builds `@{ Script = ... }` and splats it into `Invoke-Pester`, but Pester 5+ removed the legacy `-Script` parameter. This is a **real product bug** — `Get-Karma`/`Show-Karma` will throw for real users whenever the koan-execution path is actually exercised. Worth its own issue; flagging here since it's severe, but it's unrelated to this build-tooling change and I didn't touch `Invoke-Koan.ps1`.

## Acceptance criteria

- [x] Root `psakeFile.ps1` exposes `Init`, `Clean`, `Build`, `Test`, `Analyze`, and `Publish`.
- [x] `Invoke-psake ?` lists the tasks.
- [x] `Invoke-psake Test` runs the Pester suite successfully.


HeyItsGilbert:
Addressed all three findings in 16bc2c7:

1. **`build.ps1` CWD-relative paths**: `requirements.psd1` and `psakeFile.ps1` are now resolved via `Join-Path $PSScriptRoot ...` instead of bare `./...`. Verified `& 'D:/PSKoans/build.ps1' -Help` works correctly when invoked from an unrelated CWD.

2. **`exit` killing the host process**: confirmed this was real — `build.ps1` ends with `exit`, and the `BeforeDiscovery` blocks' nested `.\build.ps1 -Task Build` ran in-process, so a direct `Invoke-Pester` run on any test file (without going through the full psake pipeline first) would terminate the calling PowerShell/Pester process entirely. Fixed by running the nested build in a child `pwsh` process and checking `$LASTEXITCODE`; the parent then calls `Set-BuildEnvironment -Force` itself since the child's env var changes don't propagate back. Verified a single test file run via raw `Invoke-Pester` now completes and returns control to the caller.

3. **`$env:BHProjectPath` null in `Properties` block**: confirmed — a bare `Invoke-psake -BuildFile ./psakeFile.ps1 -TaskList ?` in a fresh session (no prior `Set-BuildEnvironment`) passed `$null` to `Join-Path` and failed before listing tasks. Switched `Test.RootDir` to resolve from `$PSScriptRoot` instead, which is always available. Verified `Invoke-psake ?` now lists all tasks with zero prior setup.

Full suite re-verified after all three fixes: 739 passed / 10 failed, same baseline as before (10 remaining failures are the pre-existing `Invoke-Koan` `-Script`/Pester 5 incompatibility called out in the PR description, unrelated to these fixes).

HeyItsGilbert:
Fixed the pre-existing \`Invoke-Koan\` bug in 54b84d4: it splatted \`Script\` (a Pester v3/v4 parameter) into \`Invoke-Pester\`, which Pester 5+ renamed to \`Path\`. This broke every real koan execution path (\`Get-Karma\`, \`Show-Karma\`) with \`ParameterBindingException\`.

Renamed \`Script\` → \`Path\` in \`Invoke-Koan.ps1\` (source + doc example), its caller in \`Get-Karma.ps1\`, and the test splats in \`Invoke-Koan.Tests.ps1\`. Also fixed a latent test bug in the same file — \`Invoke-Koan ... | Should -Not -Throw\` piped a value instead of a scriptblock into \`-Throw\`.

Full suite now **749 passed / 0 failed** (up from 739/10). \`Analyze\` still clean (pre-existing Warnings only, no Errors).

PR #500, opened 2026-09-18
build: scope changelog lint depth past pre-SemVer patch history

### LINKED ISSUE #499: Fix CHANGELOG.md to pass Keep a Changelog patch-version lint
## Goal

The reusable CI workflow added in #497 (org `powershell-ci.yml`) runs `mindsers/changelog-reader-action@v2` with `validation_level: error` against CHANGELOG.md, enforcing that patch releases (per SemVer) only contain `Fixed`/`Security` sections. This is the first time CHANGELOG.md (added in #494) has been linted, and it fails:

```
Only ''fixed, security'' sections are allowed for version 0.65.1.
Only ''fixed, security'' sections are allowed for version 0.65.2.
Only ''fixed, security'' sections are allowed for version 0.65.3.
Only ''fixed, security'' sections are allowed for version 0.66.1.
Only ''fixed, security'' sections are allowed for version 0.66.2.
0.67.1 entry is invalid.
```

Run: https://github.com/PowerShellOrg/PSKoans/actions/runs/35290428430/job/105431858210

Each flagged patch version (`0.65.1`, `0.65.2`, `0.65.3`, `0.66.1`, `0.66.2`, `0.67.1`) currently has an `### Added` section alongside/instead of `### Fixed`, which predates this org convention.

## Acceptance criteria

- Decide and apply a resolution for the flagged historical entries (e.g. re-categorize into Fixed where accurate, or document/justify an exception) without misrepresenting what actually shipped in those releases.
- `ci / Run Linters` (changelog validation step) is green in #488''s reusable CI workflow.

Blocks #488.

HeyItsGilbert:
## Summary

`mindsers/changelog-reader-action` (added via #497's reusable `powershell-ci.yml`, `validation_level: error`) validates the 10 most recent released `CHANGELOG.md` entries against Keep a Changelog / SemVer section rules, and fails:

```
Only 'fixed, security' sections are allowed for version 0.65.1.
Only 'fixed, security' sections are allowed for version 0.65.2.
Only 'fixed, security' sections are allowed for version 0.65.3.
Only 'fixed, security' sections are allowed for version 0.66.1.
Only 'fixed, security' sections are allowed for version 0.66.2.
```

## Investigation

Checked each flagged patch release against its git tag range (e.g. `git log v0.65.0..v0.65.1`) — the `Added`/`Changed` bullets are accurately attributed; those releases genuinely shipped features/refactors alongside fixes, predating this org's patch-release convention. Recategorizing them as `Fixed`-only would misrepresent what actually shipped, which the issue rules out.

## Fix

- Add `.changelog-reader.yml` setting `validation_depth: 2`, using the action's own documented escape valve to scope validation to the current + most recent released entry (both already compliant). As new releases land, the non-compliant history permanently ages out of the validated window — no historical content rewritten.
- Document the rationale in `CHANGELOG.md`'s header.

## Verification

Built `mindsers/changelog-reader-action`'s `dist/index.js` locally and ran it against this repo's `CHANGELOG.md`:
- Without the config: reproduces the exact errors from #499 (exit 1).
- With `.changelog-reader.yml`: exit 0, zero diagnostics.

Fixes #499


HeyItsGilbert:
Reverting: .changelog-reader.yml's validation_depth override doesn't actually take effect. mindsers/changelog-reader-action's action.yml declares validation_depth with a default of '10', which GitHub Actions always injects into INPUT_VALIDATION_DEPTH when the workflow doesn't set it explicitly — so the action's own `validationDepthInput || fileConfig.validation_depth` fallback always resolves to the runner-injected '10', never our config file. Confirmed by reproducing with the action's dist/index.js locally.

The 6 flagged entries roll out of the depth-10 validation window naturally as new releases land (0.65.1 in 2 releases, up through 0.66.2 in 8). 'ci / Run Linters' also isn't currently a required status check on main (branch protection still references the legacy workflow's job names), so this isn't blocking merges in the meantime. Leaving as-is per team decision; see #499.

